PR #14, opened 2026-04-08
feat: refactor QuicEndpoint API and implement mock unit tests (#4)

### LINKED ISSUE #4: Create TransportEndpoint over UDP
QUIC-based (quinn-rs)

wsnark:
Closes #4

### DIFF
--- a/src/transport/endpoint.rs
+++ b/src/transport/endpoint.rs
@@ -4,6 +4,7 @@
 use std::sync::Arc;
 
 use error_stack::{Report, ResultExt};
+use mockall_double::double;
 use quinn::{
     ClientConfig, ServerConfig, VarInt,
     crypto::rustls::{QuicClientConfig, QuicServerConfig},
@@ -15,18 +16,25 @@ use rustls::{
 
 use super::{Error, insecure_server_verifier::InsecureServerVerifier, resolver::Resolver};
 
+mod mocks;
+#[double]
+use mocks::Endpoint;
+
+pub mod connection;
+use connection::{Close, Inspect, QuicConnection};
+
 /// QUIC-based Florete Endpoint.
 #[derive(Clone)]
 pub struct QuicEndpoint {
-    endpoint: quinn::Endpoint,
+    endpoint: Endpoint,
     resolver: Arc<dyn Resolver>,
     served: Vec<String>, // For SNI validation on accept
 }
 
 // Close was caused by the endpoint, either normally or by internal error
-const ENDPOINT_CLOSE_CODE: VarInt = VarInt::from_u32(0);
+const ENDPOINT_CLOSE_CODE: u32 = 0;
 // Client behaviour was incorrect, causing the connection to close
-const CLIENT_ERROR_CODE: VarInt = VarInt::from_u32(1);
+const CLIENT_ERROR_CODE: u32 = 1;
 // Flor protocol string for ALPN
 const FLOR_ALPN: &str = "flor/1";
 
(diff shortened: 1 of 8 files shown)


src/transport/endpoint/connection.rs
```diff
@@ -0,0 +1,72 @@
...
+// Licensed under Apache-2.0 or MIT at your option.
+
+use std::any::Any;
+
+use async_trait::async_trait;
+use quinn::{ConnectionError, RecvStream, SendStream, VarInt};
+
+/// Trait to inspect connections.
+pub trait Inspect {
+    /// Get information about parameters negitiated during handshake.
```
leviathan941: negotiated?

src/transport/endpoint/connection.rs
```diff
@@ -0,0 +1,72 @@
...
+    ///
+    /// For normal quinn connections returns `Some` [`downcastable`](Box::downcast) to a
+    /// [`crypto::rustls::HandshakeData`](quinn::crypto::rustls::HandshakeData).
+    fn handshake_data(&self) -> Option<Box<dyn Any>>;
```
leviathan941: The function returns anything is not a good API.

src/transport/endpoint/mocks.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+#[cfg(test)]
+mock! {
+    pub Incoming {}
+    impl Future for Incoming {
+        type Output = Result<quinn::Connection, quinn::ConnectionError>;
+
+        fn poll<'a>(self: Pin<&mut Self>, cx: &mut Context<'a>) -> std::task::Poll<Result<quinn::Connection, quinn::ConnectionError>>;
```
leviathan941: This line is too long (135 chars).
What limit do we use?

The official guide tells it should be 100 chars:
https://doc.rust-lang.org/style-guide/#indentation-and-line-width

src/lib.rs
```diff
@@ -3,3 +3,10 @@
 
 pub mod logging;
 pub mod transport;
+
+/// Use `RUST_LOG=debug cargo test -- --nocapture` to see log output of the tests.
+#[cfg(test)]
+#[ctor::ctor]
+fn init_global_test_logging() {
```
leviathan941: Why do you initialize logs in tests with `Off` log level which means no logs will be written?

src/transport/endpoint.rs
```diff
@@ -146,49 +155,275 @@ impl QuicEndpoint {
...
+    }
+
+    #[test]
+    fn test_failure_to_create_quinn_endpoint() {
+        // Note from official Mockall example
+        // https://github.com/asomers/mockall/blob/master/mockall/examples/synchronization.rs:
+        // The mutex might be poisoned if another test fails.  But we don't
+        // care, because it doesn't hold any data. So don't unwrap the Result
+        // object; whether it's poisoned or not, we'll still hold the MutexGuard.
+        let _m = MTX.lock();
```
leviathan941: Have you tried this: https://crates.io/crates/serial_test ?

src/transport/endpoint/mocks.rs
wsnark: 100%. Auto-formatting doesn't work within macro calls

src/transport/endpoint/connection.rs
wsnark: That's original quinn's API because it allows third party crypto providers. We can improve it ofc - e.g. limiting to quinn's rustls::HandshakeData. 

```rs
fn handshake_data(&self) -> Result<quinn::crypto::rustls::HandshakeData, Error>
```
Do you think this will be better?

leviathan941: Yes, I think so

src/lib.rs
wsnark: To ensure they're off by default, can be enabled by `RUST_LOG=debug` env var only when needed

src/transport/endpoint/connection.rs
wsnark: Rewritten

leviathan941 (approved):

PR #18, opened 2026-04-15
feat: introduce QuicEndpointActor (#17)

### LINKED ISSUE #17: Transform QuicEndpoint into actor
Actor is the common pattern for components in tokio context

wsnark:
Closes #17

### DIFF
--- a/src/actor.rs
+++ b/src/actor.rs
@@ -1,0 +1,30 @@
+// Copyright (C) 2026 ReteLabs LLC.
+// Licensed under Apache-2.0 or MIT at your option.
+
+use async_trait::async_trait;
+use tokio::sync::mpsc;
+
+/// Trait for tokio actor components driven by an mpsc message channel.
+///
+/// Implement [`handle`] for actors with a single message source — the default
+/// [`run`] loop takes care of the rest.
+///
+/// Override [`run`] when the actor needs additional concurrent concerns alongside
+/// message processing, such as a background accept loop or a `watch` channel.
+#[async_trait]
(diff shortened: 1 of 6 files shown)


src/actor.rs
leviathan941, line 1: Wdyt about creating a module for internal things for such things?
Something like `utils`, but not actually.

src/transport.rs
```diff
@@ -1,15 +1,35 @@
...
+///
+/// Drop the [`QuicAcceptor`] to close and drop the endpoint.
+///
+/// This is a temporary construction path until the fundle `TransportModule` is in place.
+pub fn create_endpoint(
+    served: Vec<String>,
+    resolver: Arc<dyn resolver::Resolver>,
+    socket: std::net::UdpSocket,
+) -> Result<(QuicConnector, QuicAcceptor), Report<Error>> {
+    endpoint::QuicEndpoint::new(served, resolver, socket).map(|ep| ep.into_actor())
```
leviathan941, line 34: Are you sure we should keep the way to create QuicEndpoint without Actor wrapper?

leviathan941 (approved):

src/actor.rs
wsnark: Completely agree. Not sure about naming. Probably we can do this once we get more of such internal helpers

src/transport.rs
wsnark: QuicEndpoint is `pub(crate)`, so it isn't exposed to the app. As for merging the actor and the endpoint - I think current design is quite nice, because we test the endpoint, not the actor - which would be much harder to test probably...

PR #19, opened 2026-04-19
feat: create socks5 inbound (#2)

leviathan941:
A way to test:

1. Run Alpha node: `cargo run -- --name=Alpha`
2. Run Beta node: `cargo run -- --name=Beta`
3. Run `curl --socks5-hostname 127.0.0.1:1080 bob`

You will see GET request on "bob" side:
```
GET / HTTP/1.1
Host: bob
User-Agent: curl/8.7.1
Accept: */*
```


src/northbound/inbound/socks5.rs
```diff
@@ -0,0 +1,347 @@
...
+
+                // Reap finished tasks to keep the set bounded.
+                Some(_) = tasks.join_next() => {}
+            }
+        }
+        // Dropping JoinSet here aborts all in-flight connection tasks.
+    }
+}
+
+async fn handle_socks5(stream: TcpStream, connector: QuicConnector) -> Result<(), Report<Error>> {
```
wsnark: What is semantics for reporting errors in this function? Some conditions log errors and return Ok, some propagate. I don't understand the logic of this... need to clarify in a doc probably

src/northbound/inbound/socks5.rs
```diff
@@ -0,0 +1,347 @@
...
+    // Step 4: Open QUIC connection to the target.
+    let quic_conn = match connector.connect(&target).await {
+        Ok(conn) => conn,
+        Err(e) => {
+            log::warn!(target: LOG_TARGET, "Failed to connect to '{target}': {e:?}");
+            // Ignore errors replying here since we're already in an error path.
+            let _ = proto.reply_error(&ReplyError::HostUnreachable).await;
+            bail!(
+                e.into_report()
+                    .change_context(Error(format!("Failed to connect to '{target}'")))
```
wsnark: This produces two warn messages:
```bash
19:29:32.549  WARN socks5_inbound: Failed to connect to 'bob': Failed to connect to bob
├╴at src/core/transport/endpoint.rs:122:14
│
╰─▶ timed out
    ╰╴at src/core/transport/endpoint.rs:122:14
19:29:32.550  WARN socks5_inbound: Connection from 127.0.0.1:53774 error: Failed to connect to 'bob'
├╴at src/northbound/inbound/socks5.rs:154:22
│
├─▶ Failed to connect to bob
│   ╰╴at src/core/transport/endpoint.rs:122:14
│
╰─▶ timed out
    ╰╴at src/core/transport/endpoint.rs:122:14
```
I suggest we reduce all such warnings/errors to a single stack - to avoid excessive logging and duplication, which just complicates reading

src/northbound/inbound/socks5.rs
```diff
@@ -0,0 +1,347 @@
...
+    let client_stream = proto
+        .reply_success(bind_addr)
+        .change_context(Error("Failed to send SOCKS5 success reply".into()))
+        .await?;
+    let (mut client_read, mut client_write) = tokio::io::split(client_stream);
+
+    log::debug!(target: LOG_TARGET, "Relaying traffic for '{target}'");
+
+    tokio::select! {
+        _ = tokio::io::copy(&mut client_read, &mut quic_send) => {}
```
wsnark: It seems we need to log errors here, as DEBUG probably, just in case we need to debug connection breaks?

src/northbound/inbound/socks5.rs
```diff
@@ -0,0 +1,347 @@
...
+    async fn test_ip_target_gets_addr_type_not_supported() {
+        let addr = free_addr().await;
+        let connector = dead_connector();
+        let inbound = Socks5Inbound::new(addr, connector).await.unwrap();
+        let _handle = inbound.spawn();
+
+        let mut client = TcpStream::connect(addr).await.unwrap();
+
+        // Greeting
+        client.write_all(&[0x05, 0x01, 0x00]).await.unwrap();
```
wsnark, line 402: Here in tests we use hard-coded SOCKS5 messages for input and then parse raw output. Maybe we can use fast_socks API to work with them instead? We shouldn't test the protocol implementation here for sure...

wsnark (changes requested):
Looks very nice, just small improvements needed

src/northbound/inbound/socks5.rs
leviathan941: Actually there is no place when an error is logged but OK is returned.

For example:
```
    if cmd != Socks5Command::TCPConnect {
        log::debug!(target: LOG_TARGET, "Unsupported command {cmd:?}, rejecting");
        proto
            .reply_error(&ReplyError::CommandNotSupported)
            .change_context(Error("Failed to reply to unsupported command".into()))
            .await?;
        return Ok(());
    }
```
The log has debug level.
OK is returned because this is not an error, we just don't support anything except TCP_CONNECT.

src/northbound/inbound/socks5.rs
leviathan941: You're right, this log is redundant.

src/northbound/inbound/socks5.rs
leviathan941: Yes, I think it makes sense.

src/northbound/inbound/socks5.rs
leviathan941: ## Trade-offs:

### In favor of the client API:

* Less low-level noise, more readable
* No need to remember byte codes like ATYP, CMD, etc.

### Against:

* If the client and server come from the same library, a shared bug in fast-socks5 may go unnoticed: both sides will do the same thing incorrectly
* Error-path tests (BIND command, IPv4 address) will still have to be written manually — the client API won’t allow sending invalid/unsupported variants
* Looking at the current tests: out of five, only `test_quic_connect_failure_gets_general_failure` goes through the full domain handshake and could use the client API. The others either stop before SOCKS5 (just TCP) or intentionally send non-standard requests.

I’d keep it as is. The readability gain is minimal, while using raw bytes for error-path tests is normal and even beneficial: it documents the exact wire format the server is expected to handle.

src/northbound/inbound/socks5.rs
```diff
@@ -0,0 +1,493 @@
...
+        // New connections should be refused now
+        let result = tokio::time::timeout(Duration::from_secs(1), TcpStream::connect(addr)).await;
+        assert!(
+            result.is_err() || result.unwrap().is_err(),
+            "connections should fail after shutdown"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_happy_path_relay() {
```
leviathan941, line 460: Added the new test with successful connection. But I had to do refactoring to take `quinn::Connection` out of the equation.

It's in the last commit. Please take a look.

src/northbound/inbound/socks5.rs
```diff
@@ -1,23 +1,81 @@
...
+        (
+            Box<dyn AsyncWrite + Unpin + Send>,
+            Box<dyn AsyncRead + Unpin + Send>,
+        ),
+        BackendError,
+    >;
+}
+
+enum BackendError {
+    HostUnreachable(Report<Error>),
```
wsnark: It feels wrong to manually make nested errors. Error-stack does this for us, so you can define just flat enum (will need to derive Error for it ofc, so it can be used within the Report)

wsnark: I suggest renaming HostUnreachable to something more meaningful (we're not operating with hosts) - ConnectFailed?

src/northbound/inbound/socks5.rs
```diff
@@ -201,6 +252,49 @@ mod tests {
...
     use tokio::io::{AsyncReadExt, AsyncWriteExt};
     use tokio::net::TcpStream;
+    use tokio::sync::Mutex;
+
+    /// Mock backend that returns one side of a `tokio::io::duplex` pair.
+    ///
+    /// The test holds the other side and uses it to simulate a QUIC peer:
+    /// writes appear as `quic_recv` data (client reads them), and
+    /// data the client sends arrives as reads on the test side.
+    struct MockBackend {
```
wsnark: Suggest renaming to MockQuicBackend - to follow mockall practice of adding a prefix to a mocked type

src/core/transport/endpoint/actor.rs
```diff
@@ -24,6 +24,17 @@ pub(crate) struct ConnectMsg {
 pub struct QuicConnector(mpsc::Sender<ConnectMsg>);
 
 impl QuicConnector {
+    /// Create a connector whose actor side has already shut down.
+    ///
+    /// Every [`connect`](Self::connect) call on the returned instance will fail immediately.
+    /// Useful in tests that verify error-path behaviour without a running QUIC endpoint.
+    #[cfg(test)]
```
wsnark: Do we really need such kludges? Try mocking QuicConnector instead, it may be not that hard with mockall.

If this is really much better, then let's put this code in the end of the impl block

leviathan941: Removed

wsnark (approved):
LGTM!

PR #24, opened 2026-05-03
refactor: integrate fundle DI framework (#20)

### LINKED ISSUE #20: Integrate Fundle DI framework
https://crates.io/crates/fundle

leviathan941:


src/northbound/inbound.rs
```diff
@@ -1,4 +1,68 @@
...
+/// Construct via [`InboundBundle::try_new`].
+#[fundle::bundle]
+pub struct InboundBundle {
+    pub socks5_handle: Option<Socks5Handle>,
+}
+
+impl InboundBundle {
+    /// Build the bundle from the given dependencies.
+    ///
+    /// [Socks5Inbound] is only created if a `socks5_addr` is provided in the config.
```
wsnark: missing backticks 
```
[`Socks5Inbound`]
```

src/core/transport.rs
```diff
@@ -14,22 +14,52 @@ mod insecure_server_verifier;
...
+/// Dependencies required to construct a [`TransportBundle`].
 ///
-/// Drop the [`QuicAcceptor`] to close and drop the endpoint.
+/// This groups the transport-layer configuration values consumed by
+/// [`TransportBundle::try_new`].
+#[fundle::deps]
+pub struct TransportDeps {
+    /// Names this node serves during QUIC/TLS setup.
+    served: ServedNames,
+    /// Local UDP socket address to bind for inbound and outbound transport.
```
wsnark: Let's use ingress and egress here instead of inbound/outbound. 

Convention:
- inbound/outbound only for northbound, i.e. workload <-> vertex interactions (e.g. socks5 inbound, tcp outbound)
- ingress/egress for what goes through southbound interface - currently QUIC-encoded

src/core/transport/endpoint/actor.rs
```diff
@@ -54,6 +57,24 @@ impl QuicAcceptor {
...
+    /// Each invocation is spawned as an independent task, so connections are handled
+    /// concurrently.
+    pub fn spawn<F, Fut>(mut self, handler: F) -> QuicHandle
+    where
+        F: Fn(String, QuicConnection) -> Fut + Send + 'static,
+        Fut: Future<Output = ()> + Send + 'static,
+    {
+        let join = tokio::spawn(async move {
+            while let Some((service_name, conn)) = self.accept().await {
+                tokio::spawn(handler(service_name, conn));
```
wsnark: Shouldn't we track spawned tasks? How joining the parent task joins them too?

src/main.rs
```diff
@@ -97,43 +104,30 @@ async fn run() -> Result<(), Report<Error>> {
...
+            socks5_addr: socks5_addr.map(Socks5Addr),
+        })
+        .transport_try(|b| TransportBundle::try_new(b))
+        .change_context_lazy(bundle_err)?
+        .inbound_try_async(async |b| InboundBundle::try_new(b).await)
+        .change_context_lazy(bundle_err)
+        .await?
+        .build();
+
+    let acceptor_handle = app.transport.endpoint_acceptor.spawn(handle_connection);
```
wsnark: Let's move spawning acceptor into transport (or even the QuicEndpointActor itself) -- and reduce amount of indirection, if possible (currently endpoint->actor->QuicAcceptor->QuicHandle -- seems too much)

wsnark: But need to support 2 cases:
- full: connector + accept loop with handler
- user node: connector only (i.e. accept loop is optional, determined in config)

src/core/transport/endpoint/actor.rs
leviathan941: Good finding! Missed.

leviathan941: Returned this code to main.

src/core/transport/endpoint/actor.rs
```diff
@@ -58,45 +101,101 @@ impl QuicAcceptor {
...
+        let conflicting: Vec<&str> = msg
+            .served
+            .iter()
+            .filter(|s| self.subscribers.contains_key(s.as_str()))
+            .map(String::as_str)
+            .collect();
+
+        if !conflicting.is_empty() {
+            let _ = msg.reply.send(Err(Error(format!(
+                "services already handled by another subscriber: {}",
```
wsnark: nit: Maybe prepend with some error text? Like "Publish conflict: " or "Failed to publish: "? Or at least capitalize first letter: "Services"

src/core/transport/endpoint/actor.rs
```diff
@@ -58,45 +101,101 @@ impl QuicAcceptor {
...
+    ) {
+        let mut publish_active = true;
         loop {
             tokio::select! {
-                Some(msg) = rx.recv() => {
-                    self.handle(msg).await
+                Some(msg) = connect_rx.recv() => {
+                    self.handle_connect(msg).await;
+                },
+                msg = publish_rx.recv(), if publish_active => {
```
wsnark: I feel like this publish_active can be removed, and handling of publish_rx implemented just like connect_rx:
```
Some(msg) = publish_rx.recv() = > { 
  self.handle_publish(msg)
},
```

src/core/transport/endpoint/actor.rs
```diff
@@ -58,45 +101,101 @@ impl QuicAcceptor {
...
-    type Message = ConnectMsg;
 
-    async fn handle(&mut self, msg: ConnectMsg) {
+    async fn handle_connect(&mut self, msg: ConnectMsg) {
         let result = self.endpoint.connect(&msg.service).await;
         let _ = msg.reply.send(result);
     }
 
-    async fn run(mut self, mut rx: mpsc::Receiver<ConnectMsg>) {
+    fn handle_publish(&mut self, msg: PublishMsg) {
```
wsnark, line 206: I feel like we'd need to propagate published services to the QuicEndpoint itself somehow. This will be mandatory when mTLS is implemented -- then we'd need to propagate their certs to the ServedCertResolver (or whatever it is called).  But even now we'd rather propagate - otherwise we have to provide full list of served initially, and then publish separately -- really strange. I.e. QuicEndpoint::new() should not accept `served`, and they should be propagated by this `publish`

src/actor.rs
```diff
@@ -1,30 +0,0 @@
-// Copyright (C) 2026 ReteLabs LLC.
```
wsnark, line 1: Why it was removed? Too simple and useless for our QuicEndpointActor?

src/northbound/inbound/socks5.rs
```diff
@@ -132,9 +120,7 @@ impl Socks5Inbound {
 
     /// Start the accept loop, returning a [`Socks5Handle`] that controls its lifetime.
     pub fn spawn(self) -> Socks5Handle {
-        Socks5Handle {
-            join: tokio::spawn(self.run()),
-        }
+        Socks5Handle(LifecycleHandle::new(tokio::spawn(self.run())))
```
wsnark: Can be shortened to `Socks5Handle::new()`

src/actor.rs
leviathan941: It isn't used anymore.

src/core/transport/endpoint/actor.rs
leviathan941: Now served is passed to publish only. It's removed from the constructor.

src/core/transport/endpoint.rs
```diff
@@ -21,17 +21,25 @@ mod mocks;
...
+///
+/// Implemented by the actor layer and injected into [`QuicEndpoint`] at spawn time, keeping
+/// the endpoint decoupled from subscriber-management details.
+pub(crate) trait ServiceValidator: Send + Sync {
+    fn is_served(&self, service: &str) -> bool;
+}
+
 /// QUIC-based Florete Endpoint.
 #[derive(Clone)]
 pub(crate) struct QuicEndpoint {
```
wsnark: Now as QuicEndpoint is used only by QuicEndpointActor, which is submodule, all pub specifiers can and probably should be removed (from QuicEndpoint and its functions).

src/core/transport.rs
```diff
@@ -3,33 +3,62 @@
...
-pub fn create_endpoint(
-    served: Vec<String>,
-    resolver: Arc<dyn resolver::Resolver>,
-    socket: std::net::UdpSocket,
-) -> Result<(QuicConnector, QuicAcceptor), Report<Error>> {
-    endpoint::QuicEndpoint::new(served, resolver, socket).map(|ep| ep.into_actor())
+/// Construct via [`TransportBundle::try_new`].
+#[fundle::bundle]
+pub struct TransportBundle {
+    pub resolver: Arc<dyn resolver::Resolver>,
```
wsnark: Resolver isn't part of Transport module APIs (it is impl detail of the QuicEndpoint, as it can work with different resolvers). It should not be exposed here at all?

src/lib.rs
```diff
@@ -1,7 +1,30 @@
...
+#[derive(Debug, Clone)]
+pub struct AddrMap(pub HashMap<String, SocketAddr>);
+
+#[derive(Debug, Clone)]
+pub struct Socks5Addr(pub SocketAddr);
+
+/// Fundle DI container for the entire application configuration.
+///
+/// - `endpoint_addr`: UDP address to bind the QUIC endpoint to.
+/// - `addr_map`: maps service names to their remote UDP addresses (used by the resolver).
```
wsnark: nit: we shouldn't mention resolver in public APIs probably

src/main.rs
```diff
@@ -144,7 +148,31 @@ async fn run() -> Result<(), Report<Error>> {
...
     }
 
-    let _ = server_handle.await;
+    let endpoint_handle = app.transport.endpoint_handle;
+    let socks5_handle = app.inbound.socks5_handle;
+
+    tokio::select! {
+        result = acceptor_handle => {
+            if let Err(e) = result {
+                log::error!("Acceptor task panicked: {e:?}");
```
wsnark: panicked -> failed? for consistency in logs

src/core/transport/endpoint.rs
```diff
@@ -21,17 +21,25 @@ mod mocks;
...
+pub use actor::{QuicAcceptor, QuicConnector, QuicHandle, QuicPublisher};
 
 pub mod connection;
 use connection::{Close, Inspect, QuicConnection};
 
+/// Validates whether an incoming connection's requested service name is currently served.
+///
+/// Implemented by the actor layer and injected into [`QuicEndpoint`] at spawn time, keeping
+/// the endpoint decoupled from subscriber-management details.
+pub(crate) trait ServiceValidator: Send + Sync {
```
wsnark: Probably should be module-private (as it is used only here and in actor)

wsnark (approved):
Looks good, just small fixes needed

PR #26, opened 2026-05-16
refactor: replace std RwLock by tokio one (#20)

leviathan941:


src/core/transport/endpoint.rs
```diff
@@ -30,8 +31,9 @@ use connection::{Close, Inspect, QuicConnection};
 ///
 /// Implemented by the actor layer and injected into [`QuicEndpoint`] at spawn time, keeping
 /// the endpoint decoupled from subscriber-management details.
+#[async_trait]
 trait ServiceValidator: Send + Sync {
-    fn is_served(&self, service: &str) -> bool;
+    async fn is_served(&self, service: &str) -> bool;
```
wsnark, line 36: This is very good improvement for current code state. But once we go to implement mTLS, this validator will change into something like [ResolvesServerCert](https://docs.rs/rustls/latest/rustls/server/trait.ResolvesServerCert.html). This is sync API. And rustls suggests "async" API of [Acceptor](https://docs.rs/rustls/latest/rustls/server/struct.Acceptor.html). The problem is, Acceptor isn't really async... I hope we'd be able to wrap it in async way -- or find an existing wrapper. Maybe quinn already solves this somehow, I haven't researched further yet

wsnark: Claude suggests that ResolvesServerCert is the best way to back multiple services with one quinn's endpoint; other option is to parse client hello from Incoming manually - which doesn't sound good at all...

TLS plan (simplified):
1. Delete the ServiceValidator trait and the is_served-based Reject branch.
2. Switch SharedServiceRegistry to std::sync::RwLock<HashMap<String, ServiceEntry>>.
3. Implement rustls::server::ResolvesServerCert on SharedServiceRegistry.
4. In QuicEndpoint::new_with_abstract_socket, replace with_single_cert(...) with .with_cert_resolver(registry.clone()).
...

Given that, I'd rather keep the ServiceValidator sync (and use sync::RwLock right now)

PR #28, opened 2026-05-18
Add minimal Identities API and CLI

### LINKED ISSUE #25: Create minimal "florctl ca" and "flor id" tools to work with identities

wsnark:
Closes #25

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,0 +1,35 @@
+// Copyright (C) 2026 ReteLabs LLC.
+// Licensed under Apache-2.0 or MIT at your option.
+
+//! Shared CLI helpers used by both `flor` and `florctl` binaries.
+
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::path::Path;
+
+use error_stack::{Report, ResultExt};
+
+#[derive(Debug, thiserror::Error)]
+#[error("{0}")]
+pub struct Error(String);
+
+/// Write a file holding private key material.
+///
+/// On Unix the file is created with mode `0600` (owner read/write only). On
+/// other platforms the OS default applies and the caller should rely on
+/// directory-level protections.
+pub fn write_secret(path: &Path, bytes: &[u8]) -> Result<(), Report<Error>> {
+    let mut opts = OpenOptions::new();
+    opts.write(true).create(true).truncate(true);
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::OpenOptionsExt;
+        opts.mode(0o600);
+    }
+    let mut f = opts
+        .open(path)
+        .change_context_lazy(|| Error(format!("Failed to open {} for writing", path.display())))?;
+    f.write_all(bytes)
+        .change_context_lazy(|| Error(format!("Failed to write {}", path.display())))?;
+    Ok(())
+}
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -3,15 +3,61 @@
 
 //! Shared CLI helpers used by both `flor` and `florctl` binaries.
 
+use std::fmt;
 use std::fs::OpenOptions;
-use std::io::Write;
+use std::io::{IsTerminal, Write};
 use std::path::Path;
 
-use error_stack::{Report, ResultExt};
+use error_stack::{FrameKind, Report, ResultExt};
 
 #[derive(Debug, thiserror::Error)]
 #[error("{0}")]
 pub struct Error(String);
+
+/// Print an error report to stderr with a bold-red `error:` prefix, in the
+/// style of `cargo` / `rustc`.
+///
+/// - `verbose = false`: compact anyhow-style chain — each context's
+///   message joined by `": "`. Hides frame locations and internal lib detail.
+/// - `verbose = true`: full error-stack tree via `Debug`. Use when the
+///   compact chain doesn't point at the cause.
+pub fn print_error<E>(report: &Report<E>, verbose: bool) {
+    let prefix = error_prefix();
+    if verbose {
+        eprintln!("{prefix} {report:?}");
+    } else {
+        eprintln!("{prefix} {}", CompactChain(report));
+    }
+}
+
+fn error_prefix() -> &'static str {
+    // ANSI bold red. Suppressed when stderr isn't a TTY or NO_COLOR is set
+    // (https://no-color.org/).
+    let use_color = std::io::stderr().is_terminal() && std::env::var_os("NO_COLOR").is_none();
+    if use_color {
+        "\x1b[1;31merror:\x1b[0m"
+    } else {
+        "error:"
+    }
+}
+
+struct CompactChain<'a, E>(&'a Report<E>);
+
+impl<E> fmt::Display for CompactChain<'_, E> {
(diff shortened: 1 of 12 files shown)


src/core/identity/kind.rs
```diff
@@ -0,0 +1,313 @@
...
+
+use std::fmt;
+use std::str::FromStr;
+
+use crate::core::identity::SpiffeId;
+
+/// The class of principal a SPIFFE ID names.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Kind {
+    /// A human operator or end user.
```
leviathan941: nit: What is the difference between operator and user?

src/core/identity/kind.rs
```diff
@@ -0,0 +1,284 @@
...
+        ([_name], _) => Ok(Scope::Cluster),
+        ([node, _name], true) => Ok(Scope::Node((*node).to_string())),
+        _ => {
+            let expected = if kind.supports_node_scope() {
+                "1 (cluster) or 2 (node-scoped) trailing segments"
+            } else {
+                "1 trailing segment"
+            };
+            bail!(Error::new(format!(
+                "SPIFFE ID path {:?} has wrong shape for {kind}: expected {expected}, got {} trailing segment(s)",
```
leviathan941: nit: Isn't this line too long?

src/core/identity.rs
```diff
@@ -0,0 +1,138 @@
...
+use error_stack::{Report, ResultExt, bail};
+
+/// Build the SPIFFE ID for a principal of the given kind, name, and optional
+/// node scope. Errors if `scope_node` is set for a kind that's always
+/// cluster-scoped (see [`Kind::supports_node_scope`]).
+pub fn build_id(
+    trust_domain: &TrustDomain,
+    kind: Kind,
+    name: &str,
+    scope_node: Option<&str>,
```
leviathan941: I don't like API that let you screw things up.

I believe that if an API does not allow combinations of any parameters, then there should be no way to pass them.

May be Kind should be split up to two enums?

For example:
```rust
pub enum ClusterKind {
    User,
    Node,
    ControlPlane,
    ManagementPlane,
}

pub enum NodeKind {
    Service,
    Vertex,
}

pub enum Kind {
    Cluster(ClusterKind),
    Node(NodeKind),
}

pub fn build_cluster_id(
    trust_domain: &TrustDomain,
    kind: ClusterKind,
    name: &str,
) -> Result<SpiffeId, Report<Error>> {
    let segments = vec![kind.as_segment(), name];

    SpiffeId::from_segments(trust_domain.clone(), &segments)
        .change_context(Error::new("Failed to construct SPIFFE ID"))
}

pub fn build_node_id(
    trust_domain: &TrustDomain,
    kind: NodeKind,
    node: &str,
    name: &str,
) -> Result<SpiffeId, Report<Error>> {
    let segments = vec![kind.as_segment(), node, name];

    SpiffeId::from_segments(trust_domain.clone(), &segments)
        .change_context(Error::new("Failed to construct SPIFFE ID"))
}
```

src/cli.rs
```diff
@@ -0,0 +1,81 @@
...
+        eprintln!("{prefix} {}", CompactChain(report));
+    }
+}
+
+fn error_prefix() -> &'static str {
+    // ANSI bold red. Suppressed when stderr isn't a TTY or NO_COLOR is set
+    // (https://no-color.org/).
+    let use_color = std::io::stderr().is_terminal() && std::env::var_os("NO_COLOR").is_none();
+    if use_color {
+        "\x1b[1;31merror:\x1b[0m"
```
leviathan941: nit: Why don't use style::AnsiColor for coloring?
We already use it in the logger.

src/main.rs
```diff
@@ -38,19 +85,67 @@ struct AppBundle {
...
-        std::process::exit(1);
+fn main() {
+    let args = Args::parse();
+    let verbose = args.verbose;
+    match args.cmd {
+        Cmd::Id {
+            action: IdAction::Keygen { .. },
+        } => {
+            // Synchronous path — no tokio needed.
+            if let Err(e) = run_id_keygen(args.cmd) {
```
leviathan941: Why do not pass IdAction::Keygen to run_id_keygen?
You have already know that cmd is IdAction::Keygen there.

src/main.rs
```diff
@@ -38,19 +85,67 @@ struct AppBundle {
...
+    let id = build_id(&td, kind, &name, scope.as_deref())
+        .change_context(Error("Failed to build SPIFFE ID".into()))?;
+    let (key, csr_pem) = keygen_csr(&id).change_context(Error("Keygen failed".into()))?;
+
+    write_secret(&out_key, key.serialize_pem().as_bytes())
+        .change_context(Error("Failed to write private key".into()))?;
+    std::fs::write(&out_csr, csr_pem.as_bytes())
+        .change_context_lazy(|| Error(format!("Failed to write CSR to {}", out_csr.display())))?;
+
+    println!("SPIFFE ID: {id}");
```
leviathan941, line 127: Why do you use println here instead of our logger?

leviathan941 (changes requested):

src/main.rs
```diff
@@ -38,19 +85,67 @@ struct AppBundle {
...
+        } => {
+            // Synchronous path — no tokio needed.
+            if let Err(e) = run_id_keygen(args.cmd) {
+                print_error(&e, verbose);
+                std::process::exit(1);
+            }
+        }
+        Cmd::Demo { .. } => {
+            logging::logger::init(log::LevelFilter::Info).expect("Failed to initialize logger");
+            let rt = tokio::runtime::Runtime::new().expect("Failed to build tokio runtime");
```
leviathan941: nit:
```rust
tokio::runtime::Builder::new_multi_thread()
    .enable_all()
    .build()?
```

src/core/identity/kind.rs
wsnark: Good catch: no difference from library PoV; operator is user with a special role that has write access to config-server - but that's irrelevant to identities module

src/core/identity.rs
wsnark: Service (and Vertex) can be cluster-scoped as well (load-balanced), so this approach isn't working. But I've come with a different solution, which keeps Kind and Scope as separate axes, yet type-safe about allowed combinations

src/main.rs
wsnark: Because this is output to the user, not a log entry

src/cli.rs
wsnark: used the one from Clap (it also exposes it)

src/core/identity/kind.rs
```diff
@@ -51,11 +61,41 @@ impl Kind {
...
+            NodeScopableKind::Vertex => Self::Vertex,
+        }
+    }
+}
+
+impl NodeScopableKind {
+    pub const fn as_segment(self) -> &'static str {
+        match self {
+            Self::Service => "service",
+            Self::Vertex => "vertex",
```
leviathan941: nit:
```rust
            Self::Service => Kind::Service.as_segment(),
            Self::Vertex => Kind::Vertex.as_segment(),
```

leviathan941 (approved):

PR #29, opened 2026-05-23
feat: create TCP direct outbound component (#3)

leviathan941:
## How to test

1. Run `nc` utility: `nc -lk 127.0.0.1 32440`.
2. Run flor demo at Delta config: `cargo run --bin flor -- demo --name Delta`.
3. Run flor at Alpha config: `cargo run --bin flor -- demo --name Alpha`.
4. Wait for the message on the nc side: `Hello from eve`.

Refs: #3

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,14 +22,17 @@ pub struct Socks5Addr(pub SocketAddr);
 pub struct TcpDirectTargets(pub HashMap<String, SocketAddr>);
 
 /// Fundle DI container for the entire application configuration.
-///
-/// - `endpoint_addr`: UDP address to bind the QUIC endpoint to.
-/// - `addr_map`: maps service names to their remote UDP addresses.
-/// - `socks5_addr`: if present, the local address to bind the SOCKS5 listener to.
 #[fundle::bundle]
 pub struct AppConfigBundle {
+    /// The UDP address to bind the QUIC endpoint to.
     pub endpoint_addr: EndpointAddr,
+
+    /// A mapping of service names to their remote UDP addresses.
     pub addr_map: AddrMap,
+
+    /// If present, the local address to bind the SOCKS5 inbound listener to.
     pub socks5_addr: Option<Socks5Addr>,
+
+    /// A mapping of service names to their remote TCP addresses for TCP direct outbound.
     pub tcp_direct_targets: TcpDirectTargets,
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,6 +18,9 @@ pub struct AddrMap(pub HashMap<String, SocketAddr>);
 #[derive(Debug, Clone)]
 pub struct Socks5Addr(pub SocketAddr);
 
+#[derive(Debug, Clone)]
+pub struct TcpDirectTargets(pub HashMap<String, SocketAddr>);
+
 /// Fundle DI container for the entire application configuration.
 ///
 /// - `endpoint_addr`: UDP address to bind the QUIC endpoint to.
@@ -28,4 +31,5 @@ pub struct AppConfigBundle {
     pub endpoint_addr: EndpointAddr,
     pub addr_map: AddrMap,
     pub socks5_addr: Option<Socks5Addr>,
+    pub tcp_direct_targets: TcpDirectTargets,
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,17 +7,20 @@ use clap::{Args as ClapArgs, Parser, Subcommand};
 use error_stack::{Report, ResultExt};
 
 use flor::{
-    AddrMap, AppConfigBundle, EndpointAddr, Socks5Addr,
+    AddrMap, AppConfigBundle, EndpointAddr, Socks5Addr, TcpDirectTargets,
     cli::{print_error, write_secret},
     core::{
         identity::{Kind, TrustDomain, build_id, keygen_csr},
         transport::{
-            QuicConnector, TransportBundle,
+            QuicConnector, QuicPublisher, TransportBundle,
             endpoint::connection::{Accept, Open},
         },
     },
     logging,
-    northbound::inbound::{Error as InboundError, InboundBundle},
+    northbound::{
+        inbound::{Error as InboundError, InboundBundle},
+        outbound::{Error as OutboundError, OutboundBundle},
+    },
     utils::report::ErrorReport,
 };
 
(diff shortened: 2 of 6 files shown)


src/main.rs
```diff
@@ -164,14 +171,31 @@ async fn demo_main(node_name: String) -> Result<(), Report<Error>> {
...
+                )]),
             ),
         ),
     ]);
     // Directed workload connections: (client, server)
-    let conn_list = vec![("alice", "bob"), ("carol", "bob")];
+    let conn_list = vec![
+        ("alice", "bob"),
+        ("carol", "bob"),
+        ("eve", "tcp-echo.demo.flor.local"),
```
wsnark, line 194: let's use consistent naming scheme in all examples. Two options for this PR:
1. Adjust to previous pattern, then "tcp-echo.demo.flor.local" -> "dave" or something like this
2. Adjust to our production pattern for services: `<svc>.<node>.<cluster>.rete`, then it would be `tcp-echo.delta.rete-lovers.rete`  

We used `rete-lovers` cluster name in the docs, maybe should rename to avoid double rete. What about `tcp-echo.delta.orion.rete` - sounds aligned with our alpha, beta, gamma for nodes. Or `tcp-echo.delta.demo-cluster.rete`

src/main.rs
```diff
@@ -139,15 +143,17 @@ fn run_demo(node_name: String) -> Result<(), Report<Error>> {
 }
 
 async fn demo_main(node_name: String) -> Result<(), Report<Error>> {
-    // Initial version uses workload names instead of identities.
-    // Node configuration: (quic_addr, served_workloads, socks5_addr)
+    // Initial demo services use workload names; TCP direct services use DNS-style names.
+    // The transport currently uses service names directly as QUIC TLS SNI values.
+    // Node configuration: (quic_addr, demo_served_workloads, socks5_addr, tcp_direct_targets)
     let service_map = HashMap::from([
```
wsnark, line 149: Let's use configuration scheme that is close to our C0 target. It looks like:
```json5
[
 "Alpha":  // node name
  { 
    "listen": "127.0.0.1:31337", // where QUIC endpoint of the node listens
    "workloads":  [ // array of workloads on the node
       {
        "name": "alice", // alice is a user, not service -- will be differentiated when SPIFFE IDs and mTLS land
        "io": [ { "kind": "socks5", "listen": "127.0.0.1:18000", "direction": "inbound"  } ]
       },
       {
         "name": "tcp-echo.alpha.orion.rete", // this is a true service, hence it has DNS name
         "io": [ { "kind": "tcp", "upstream": "127.0.0.1:32440", "direction": "outbound" } ]
       }
     ] 
   }
]
```
For now, we shouldn't parse json, so we can represent it like: `(quic_addr, socks5_inbounds, tcp_outbounds)`, where `socks5_inbounds: vec![(name, listen_addr)]`, `tcp_outbounds: vec![(service_name, upstream_addr)]`

wsnark: If we want to save our demo outbounds (which I doubt), then it would be something like `native_outbounds: vec![(service_name)]`

wsnark (changes requested):

src/main.rs
leviathan941: Made the changes in the separate PR: https://github.com/rete-labs/flor/pull/34
PTAL

src/northbound/outbound/tcp.rs
```diff
@@ -0,0 +1,349 @@
...
+    let Some(target_addr) = targets.get(&service_name).copied() else {
+        log::debug!(target: LOG_TARGET, "No TCP target configured for '{service_name}'");
+        return;
+    };
+
+    let mut streams = JoinSet::new();
+
+    loop {
+        tokio::select! {
+            stream = conn.accept_stream() => match stream {
```
wsnark, line 108: Nice feature: we can map multiple TCP connections to QUIC streams, instead of establishing full QUIC connection per TCP connection. It hasn't been well-thought though -- e.g. SOCK5 inbound currently doesn't support this. Also we may want to keep our QUIC connection alive a bit, in order to optimize multiple req/resp (e.g. HTTP) without keeping TCP connections alive.

Created #36 to track this feature.

src/northbound/outbound/tcp.rs
```diff
@@ -0,0 +1,349 @@
...
+
+impl<T> QuicStream for T where T: AsyncRead + AsyncWrite + Unpin + Send {}
+
+#[async_trait]
+trait QuicInboundConnection: Send + Sync {
+    async fn accept_stream(&self) -> Result<Box<dyn QuicStream>, Report<Error>>;
+}
+
+#[async_trait]
+impl QuicInboundConnection for QuicConnection {
```
wsnark: These Quic- abstractions may be useful for other outbounds/inbounds, shouldn't we put them in QuicEndpoint API (connection.rs)?

src/lib.rs
wsnark: Missing docstring for new field; maybe move them inside the struct, instead of listing here?

src/northbound/outbound.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright (C) 2026 ReteLabs LLC.
+// Licensed under Apache-2.0 or MIT at your option.
+
+use error_stack::ResultExt;
+
+use crate::{
+    TcpDirectTargets,
```
wsnark, line 9: I feel uncomfortable when a module depends on something defined in lib.rs. Shouldn't we define this TcpDirectTargets here or even in tcp.rs, and expose to the lib.rs? (Same for other fields there, I see we have current design as a pattern, not something new added in this commit)

src/northbound/outbound.rs
```diff
@@ -0,0 +1,65 @@
...
+        .keys()
+        .cloned()
+        .collect::<Vec<_>>()
+        .join(", ");
+    let tcp_direct = TcpDirectOutbound::new(deps.tcp_direct_targets.0, deps.quic_publisher)
+        .await
+        .change_context(Error("Failed to create TCP direct outbound".into()))?
+        .spawn();
+
+    log::info!("TCP direct outbound serving: {services}");
```
wsnark: It would be nice to log the map (service -> tcp address), because currently tcp targets are never logged (unlike SOCKS5, where both name and listening address are logged)

src/northbound/outbound.rs
leviathan941: These things are just aliases for passing configure parameters from outside of the library to its components.
It's similar to command line parameters or something like that.
So, I think they must be where they are.

src/northbound/outbound/tcp.rs
leviathan941: Moved into the outbound module.

wsnark (approved):
LGTM

PR #38, opened 2026-06-18
feat: implement mTLS support (#32)

### LINKED ISSUE #32: Implement mTLS support
Follow ADR-0006 to implement mTLS in flor

wsnark:
Closes #32


src/core/identity/dialable.rs
```diff
@@ -0,0 +1,341 @@
...
+        Ok(Self(id))
+    }
+
+    /// The wrapped canonical identity.
+    pub fn id(&self) -> &SpiffeId {
+        &self.0
+    }
+
+    /// Consume into the wrapped [`SpiffeId`].
+    pub fn into_id(self) -> SpiffeId {
```
leviathan941: Is this function really necessary?
If you need to extract SpiffeId from here, you just can take the reference via `id` and clone it.

src/core/identity/dialable.rs
```diff
@@ -0,0 +1,341 @@
...
+        self.0
+    }
+
+    /// The dialable kind (`Service` or `Vertex`).
+    pub fn kind(&self) -> Kind {
+        kind_of(&self.0).expect("Dialable validates kind at construction")
```
leviathan941: Perhaps, it would be better to store Kind inside the structure? Then we don't have to use expect here.

src/core/identity/dialable.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    /// Synthesise the `.rete` convenience hostname. Infallible: a `Dialable` is
+    /// always a well-formed service/vertex identity. Trust domains may be dotted.
+    pub fn render(&self) -> String {
+        let td = self.0.trust_domain().as_str();
+        let name = leaf_name(&self.0).expect("Dialable always has a name segment");
```
leviathan941: If this thing is mandatory why it isn't checked inside the constructor?
I think that we must check everything in the constructor and store all things already parsed and checked to prevent:
- double validation. It's not so bad, but just looks unnecessary in general.
- possible failures in the future usage of the created instance. If an instance is created it must be valid.

leviathan941 (changes requested):

src/core/identity/dialable.rs
wsnark: Fixed

src/core/identity/dialable.rs
wsnark: Removed

src/northbound/inbound/socks5.rs
```diff
@@ -136,43 +141,45 @@ impl Socks5Inbound {
...
-                    let service_name = service_name.clone();
+                    let caller = caller.clone();
+                    let principal = principal.clone();
                     tasks.spawn(async move {
-                        if let Err(e) = handle_socks5(stream, backend).await {
+                        if let Err(e) = handle_socks5(stream, backend, caller).await {
                             log::warn!(target: LOG_TARGET,
-                                "Connection from {peer_addr} via SOCKS5 service '{service_name}'
-                                error: {e:?}");
+                                "Connection from {peer_addr} via SOCKS5 for principal '{principal}' error: {e:?}");
```
leviathan941: This line is longer than 100 chars

src/core/transport/endpoint.rs
```diff
@@ -167,23 +251,48 @@ impl QuicEndpoint {
...
+                Err(e) => {
+                    accepted.close(ENDPOINT_CLOSE_CODE, b"internal-error");
+                    return Err(e.change_context(Error(format!(
+                        "Failed to resolve peer identity for connection to '{target}'"
+                    ))));
                 }
+            };
+            log::debug!("Accepted connection for '{target}' from peer '{peer_id}'");
+
+            let conn = accepted.into_connection(peer_id);
```
leviathan941: What if I pass peer_id from another connection here?
I don't think we should allow this behavior.

Why don't put the code of getting peer_id inside into_connection and remove the parameter?

src/core/transport/endpoint/registry.rs
```diff
@@ -0,0 +1,205 @@
...
+use crate::core::identity::SpiffeId;
+
+use super::connection::QuicConnection;
+use super::verifier::ServerCertRegistry;
+
+/// Channel delivering accepted connections to a subscriber, each tagged with the
+/// target identity it was dialed against.
+pub(super) type Dispatch = mpsc::Sender<(SpiffeId, QuicConnection)>;
+
+/// One published service: its identity, the cert the server presents for it, and
```
leviathan941, line 27: nit: Is not "principal" more accurate than "service" here?

scripts/e2e-relay.sh
```diff
@@ -0,0 +1,124 @@
...
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT"
+
+readonly SOCKS5_ADDR="127.0.0.1:1080"
+readonly ECHO_ADDR="127.0.0.1:32450"
+readonly TARGET_HOST="tcp-echo.beta.demo.flor.rete"
+
+command -v python3 >/dev/null || {
+  echo "error: python3 is required (used for the echo upstream + SOCKS5 client)" >&2
```
leviathan941, line 26: nit: If python3 is already used anyway, why don't write the whole script in it?

src/core/transport/endpoint.rs
```diff
@@ -24,46 +25,103 @@ pub mod actor;
...
+    Closed,
+}
+
+/// The SNI routing label for a dial target: the transport's SNI-derivation seam.
+/// Today the rendered `.rete` name; the receiver resolves it by registry lookup,
+/// never by parsing an identity out of it (ADR-0007). Used by both the dialing
+/// side (`connect`) and the publishing side (the registry key).
+fn sni_for(target: &Dialable) -> String {
+    target.render()
+}
```
leviathan941, line 85: Very strange function. It adds nothing to the `render` function. What's the point of it?
Are you going to change the implementation in the future?

src/main.rs
```diff
@@ -46,7 +51,7 @@ enum Cmd {
     /// Run the demo (legacy, replaced by `agent run` once the daemon lands).
     Demo {
         /// Select node config.
-        #[arg(long, default_value = "Alpha", value_parser = ["Alpha", "Beta"])]
+        #[arg(long, default_value = "alpha", value_parser = ["alpha", "beta"])]
```
leviathan941, line 54: Is it really necessary change?

src/main.rs
```diff
@@ -139,81 +144,140 @@ fn run_demo(node_name: String) -> Result<(), Report<Error>> {
...
+}
+
+/// A demo node's declarative config: where its QUIC endpoint binds, the SOCKS5
+/// principals it proxies for, and the TCP-direct services it serves.
+struct NodeConfig {
+    quic_addr: SocketAddr,
+    /// SOCKS5 caller principals: `(name, local listen address)`.
+    socks5: Vec<(&'static str, SocketAddr)>,
+    /// TCP-direct services: `(name, local upstream address)`.
+    services: Vec<(&'static str, SocketAddr)>,
```
leviathan941, line 189: Why just `services`? It's not enough distinctive I think.
May be `tcp_outbounds` and `socks5_inbounds`?

src/main.rs
```diff
@@ -139,81 +144,140 @@ fn run_demo(node_name: String) -> Result<(), Report<Error>> {
...
+}
+
+fn sock(s: &str) -> SocketAddr {
```
leviathan941, line 192: Why do we really need this function? With `expect` in implementation.

src/core/transport/endpoint/registry.rs
wsnark: Well, principal is more general. I.e. each service is a principal. This entry is a service though, specifically. So we can and should call it as such here, I think

scripts/e2e-relay.sh
wsnark: Will rework this script within https://github.com/rete-labs/flor/pull/41

src/core/transport/endpoint.rs
wsnark: This function serves very important role - hiding the implementation detail from the caller. Nothing depends on the fact it is just render() now -- so we can replace the mechanism to something more complex if needed. No plans to do so in C0/C1 though.

src/main.rs
wsnark: This code is removed in the follow-up PR https://github.com/rete-labs/flor/pull/41 - so let's keep it as is here

src/main.rs
wsnark: Just a helper to parse hardcoded sockaddrs. Will be removed in https://github.com/rete-labs/flor/pull/41, so let's keep for now

src/main.rs
wsnark: This is about case-sensitivity in node names. Very important, to be decided in #105. Let's use lowercase for now (and this code is removed shortly)

src/core/transport/endpoint.rs
wsnark: Refactored

src/core/transport/endpoint.rs
```diff
@@ -62,6 +62,20 @@ enum AcceptOutcome {
...
+/// A routine dial failure is the caller's to handle, whereas a fault means
+/// the endpoint itself is broken and must be shut down.
+enum ConnectError {
+    /// A routine, per-call failure — unresolvable target, unreachable peer, or a
+    /// handshake the verifiers rejected. Reported to the dialing caller; the
+    /// endpoint should be kept up.
+    Failed(Report<Error>),
+    /// An internal inconsistency: the endpoint's invariant is broken, so it must
+    /// be shut down.
+    Fault(Report<Error>),
```
leviathan941: `Failed` and `Fault` are too similar names.

I propose one of the following options:
- Operational/Internal
- External/Internal

leviathan941 (approved):

src/core/transport/endpoint.rs
wsnark: Agree, used External/Internal

Issue #30, opened 2026-05-25
Implement `retectl validate` for rete configs


wsnark:
Validator rules to implement: https://florete.tech/docs/implementation/c0-tended-tunnels/mgmt-plane/validate-and-compile/#validator-rules

Except rule 4 (enrollment) - let's skip it for C0.

Rete configs are described in https://florete.tech/docs/implementation/c0-tended-tunnels/mgmt-plane/source-layout/

PR #40, opened 2026-06-24
feat: add rete config validator (#30)

leviathan941:
Refs #30

### DIFF
--- a/src/bin/retectl.rs
+++ b/src/bin/retectl.rs
@@ -1,7 +1,7 @@
 // Copyright (C) 2026 ReteLabs LLC.
 // Licensed under Apache-2.0 or MIT at your option.
 
-//! `florctl` — operator-only rete-authoring CLI.
+//! `retectl` — operator-only rete-authoring CLI.
 //!
 //! The CA private key is only ever touched by this binary; it never ships to a node.
 
@@ -12,7 +12,8 @@ use clap::{Args, Parser, Subcommand};
 use error_stack::{Report, ResultExt};
 
 use flor::{
-    cli::{print_error, write_secret},
+    cli::{print_error, print_load_error, write_secret},
+    config::rete::{LoadOpts, load, validate},
     core::identity::{Ca, Kind, TrustDomain, build_id},
 };
 
@@ -21,9 +22,9 @@ use flor::{
 struct Error(String);
 
 #[derive(Parser, Debug)]
-#[command(name = "florctl", about = "Florete operator CLI (rete authoring)")]
+#[command(name = "retectl", about = "Florete operator CLI (rete authoring)")]
 struct Cli {
-    /// Show the full error-stack chain on failure (default: compact `: `-joined chain).
+    /// Enable debug logging (also prints full error chains on failure).
     #[arg(short, long, global = true)]
     verbose: bool,
 
@@ -38,6 +39,19 @@ enum Cmd {
         #[command(subcommand)]
         action: CaAction,
     },
+    /// Validate rete config: schema, cross-references, access consistency.
+    Validate(ValidateArgs),
+}
+
+#[derive(Args, Debug)]
+struct ValidateArgs {
+    /// Repository root directory.
+    #[arg(long, default_value = ".")]
+    repo: PathBuf,
+
+    /// Override file discovery with explicit paths or globs (repeatable).
+    #[arg(short, long = "file", value_name = "FILE")]
+    files: Vec<String>,
 }
 
 #[derive(Subcommand, Debug)]
@@ -97,6 +111,14 @@ struct CaSignArgs {
 
 fn main() {
     let cli = Cli::parse();
+
+    let log_level = if cli.verbose {
+        log::LevelFilter::Debug
+    } else {
+        log::LevelFilter::Warn
+    };
+    let _ = flor::logging::logger::init(log_level);
+
     if let Err(e) = run(cli.cmd) {
         print_error(&e, cli.verbose);
         std::process::exit(1);
@@ -109,7 +131,34 @@ fn run(cmd: Cmd) -> Result<(), Report<Error>> {
             CaAction::Init(args) => ca_init(args),
             CaAction::Sign(args) => ca_sign(args),
         },
+        Cmd::Validate(args) => cmd_validate(args),
     }
+}
+
+fn cmd_validate(args: ValidateArgs) -> Result<(), Report<Error>> {
+    let opts = LoadOpts {
+        repo: args.repo,
+        files: args.files,
+    };
+
+    let model = load(&opts).map_err(|e| {
+        print_load_error(&e);
+        Report::new(Error("config load failed".into()))
+    })?;
+
+    let violations = validate(&model);
+
+    if violations.is_empty() {
+        println!("ok");
+    } else {
+        for v in &violations {
+            eprintln!("  [{}] {}", v.rule, v.message);
+        }
+        eprintln!("error: {} violation(s) found", violations.len());
+        std::process::exit(1);
+    }
+
+    Ok(())
 }
 
 fn ca_init(args: CaInitArgs) -> Result<(), Report<Error>> {
--- a/src/bin/retectl.rs
+++ b/src/bin/retectl.rs
@@ -12,7 +12,7 @@ use clap::{Args, Parser, Subcommand};
 use error_stack::{Report, ResultExt};
 
 use flor::{
-    cli::{print_error, print_load_error, write_secret},
+    cli::{print_error, write_secret},
     config::rete::{LoadOpts, load, validate},
     core::identity::{Ca, Kind, TrustDomain, build_id},
 };
@@ -119,30 +119,30 @@ fn main() {
     };
     let _ = flor::logging::logger::init(log_level);
 
-    if let Err(e) = run(cli.cmd) {
+    if let Err(e) = run(cli.cmd, cli.verbose) {
         print_error(&e, cli.verbose);
         std::process::exit(1);
     }
 }
 
-fn run(cmd: Cmd) -> Result<(), Report<Error>> {
+fn run(cmd: Cmd, verbose: bool) -> Result<(), Report<Error>> {
     match cmd {
         Cmd::Ca { action } => match action {
             CaAction::Init(args) => ca_init(args),
             CaAction::Sign(args) => ca_sign(args),
         },
-        Cmd::Validate(args) => cmd_validate(args),
+        Cmd::Validate(args) => cmd_validate(args, verbose),
     }
 }
 
-fn cmd_validate(args: ValidateArgs) -> Result<(), Report<Error>> {
+fn cmd_validate(args: ValidateArgs, verbose: bool) -> Result<(), Report<Error>> {
     let opts = LoadOpts {
         repo: args.repo,
         files: args.files,
     };
 
     let model = load(&opts).map_err(|e| {
-        print_load_error(&e);
+        print_error(&e, verbose);
         Report::new(Error("config load failed".into()))
     })?;
 
--- a/src/bin/retectl.rs
+++ b/src/bin/retectl.rs
@@ -12,7 +12,7 @@ use clap::{Args, Parser, Subcommand};
 use error_stack::{Report, ResultExt};
 
 use flor::{
-    cli::{print_error, write_secret},
+    cli::{compact_chain, print_error, print_error_lines, write_secret},
     config::rete::{LoadOpts, load, validate},
     core::identity::{Ca, Kind, TrustDomain, build_id},
 };
@@ -24,10 +24,6 @@ struct Error(String);
 #[derive(Parser, Debug)]
 #[command(name = "retectl", about = "Florete operator CLI (rete authoring)")]
 struct Cli {
-    /// Enable debug logging (also prints full error chains on failure).
-    #[arg(short, long, global = true)]
-    verbose: bool,
-
     #[command(subcommand)]
     cmd: Cmd,
 }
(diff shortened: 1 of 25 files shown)


src/bin/retectl.rs
```diff
@@ -109,7 +131,35 @@ fn run(cmd: Cmd) -> Result<(), Report<Error>> {
...
+
+fn cmd_validate(args: ValidateArgs) -> Result<(), Report<Error>> {
+    let opts = LoadOpts {
+        repo: args.repo,
+        files: args.files,
+    };
+
+    let model = load(&opts).map_err(|e| {
+        print_load_error(&e);
+        Report::new(Error("config load failed".into()))
```
wsnark: This causes 2 `error` outputs on each error, that seem misleading. Shouldn't we write single red `error` for a failure (and if we see multiple, maybe can report all of them -- but for now we fail fast, on first one...)

Image

src/cli.rs
```diff
@@ -61,6 +61,17 @@ impl<E> fmt::Display for CompactChain<'_, E> {
     }
 }
 
+/// Print a load error to stderr (clean readable message) and to the log (full chain).
+///
+/// Use this for operational failures during config loading — missing files, I/O errors,
+/// parse failures — where the full error-stack debug output is developer-relevant but
+/// not useful to the operator reading the terminal.
+pub fn print_load_error<E: std::error::Error>(report: &Report<E>) {
```
wsnark: It is strange to call this `print_load_error` - I'd rather call it per what it does, not where it can be used (on config loading) - as it is quite general

src/bin/retectl.rs
```diff
@@ -109,7 +131,35 @@ fn run(cmd: Cmd) -> Result<(), Report<Error>> {
...
+}
+
+fn cmd_validate(args: ValidateArgs) -> Result<(), Report<Error>> {
+    let opts = LoadOpts {
+        repo: args.repo,
+        files: args.files,
+    };
+
+    let model = load(&opts).map_err(|e| {
+        print_load_error(&e);
```
wsnark: Currently we don't have persistent logs functionality, only stderr, so this function causes duplicate output of the errors (full + compact). 

I suggest we drop logging functionality from retectl for now, and introduce it together with persistent logger (syslog/journalctl/file - whatever, need to design this properly + need to take in account a possible feature to upload logs to us - sort of bug reports?)

Image

src/bin/retectl.rs
```diff
@@ -21,9 +22,9 @@ use flor::{
 struct Error(String);
 
 #[derive(Parser, Debug)]
-#[command(name = "florctl", about = "Florete operator CLI (cluster authoring)")]
+#[command(name = "retectl", about = "Florete operator CLI (cluster authoring)")]
 struct Cli {
-    /// Show the full error-stack chain on failure (default: compact `: `-joined chain).
+    /// Enable debug logging (also prints full error chains on failure).
```
wsnark: That feels a bit strange; I think `--verbose` isn't about debugging the `retectl`, it should help "debugging" the rete configs. Debug logs can be enabled via RUST_LOG now... And later we need to design proper UX for collecting logs to file bug reports (certainly not within this PR...)

src/config/rete/loader.rs
```diff
@@ -0,0 +1,151 @@
...
+            write!(f, "\n  {}", e)?;
+        }
+        Ok(())
+    }
+}
+
+impl std::error::Error for ParseFailures {}
+
+#[derive(Debug, thiserror::Error)]
+pub enum LoadError {
```
wsnark, line 31: Contents of this enum is strange: named LoadError, but also contains Discovery, Parse and Merge errors, which have separate modules - maybe they should have their own errors? Or, if they all are part of the Load process, then maybe their modules should be submodules? Unclear, but current state feels wrong...

src/config/rete/loader.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[error("'rete.yaml' not found at repo root '{}'", .0.display())]
+    MissingAnchor(PathBuf),
```
wsnark: Let's not call this Anchor, it is a special term in YAML, and so this is misleading. I know it is called as such in the doc, but we maybe even should fix the doc - or maybe keep it, as it is in different context...

src/bin/retectl.rs
```diff
@@ -38,6 +39,19 @@ enum Cmd {
...
     },
+    /// Validate rete config: schema, cross-references, access consistency.
+    Validate(ValidateArgs),
+}
+
+#[derive(Args, Debug)]
+struct ValidateArgs {
+    /// Repository root directory.
+    #[arg(long, default_value = ".")]
+    repo: PathBuf,
```
wsnark, line 46: Currently usage of `retectl validate` isn't ergonomic enough: if there is no `repo` in the `.`, the error message is unhelpful - user has to call `retectl validate --help` to learn that key is named `--repo`, otherwise there is no hints.

Image

src/config/rete/source.rs
```diff
@@ -0,0 +1,130 @@
+// Copyright (C) 2026 ReteLabs LLC.
+// Licensed under Apache-2.0 or MIT at your option.
+
+//! Source file discovery for rete config repositories.
```
wsnark: Maybe rename source.rs into discover.rs, similar to merge and validate?

src/config/rete/source.rs
```diff
@@ -0,0 +1,130 @@
...
+use std::path::{Path, PathBuf};
+
+use glob::glob;
+
+use super::loader::LoadError;
+use super::model::Source;
+use error_stack::{Report, ResultExt};
+
+/// Options controlling where config files are loaded from.
+pub struct LoadOpts {
```
wsnark: It is strange that LoadOpts are located not in `loader.rs`, but in `source.rs`...

src/config/rete/source.rs
```diff
@@ -0,0 +1,130 @@
...
+        return expand_overrides(&opts.files);
+    }
+    discover_from_repo(&opts.repo, source)
+}
+
+fn discover_from_repo(
+    repo: &Path,
+    source: Option<&Source>,
+) -> Result<Vec<PathBuf>, Report<LoadError>> {
+    // rete.yaml is handled separately by the loader (already parsed in Phase 1);
```
wsnark: Maybe we shouldn't refer to impl details of the other module ("Phase 1" here) ? Though the logic itself is valid, of course

src/config/rete/source.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    let include_globs: Vec<String> = source
+        .and_then(|s| s.include.clone())
+        .unwrap_or_else(|| vec!["**/*.yaml".into(), "**/*.yml".into()]);
+
+    let exclude_globs: Vec<String> = source.and_then(|s| s.exclude.clone()).unwrap_or_default();
+
+    for pattern in &include_globs {
+        let full_pattern = repo.join(pattern);
+        let full_pattern_str = full_pattern.to_string_lossy();
```
wsnark: Why lossy? maybe we should not allow non-UTF-8 symbols in the patterns? Or what is the best practice to handle non-UTF-8 in filenames/patterns?

src/config/rete/loader/discover.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub repo: PathBuf,
+    /// When non-empty, overrides discovery entirely.
+    /// At least one selected file must contain the `rete` block.
+    pub files: Vec<String>,
+}
+
+/// Discover all YAML source files to load.
+///
+/// Returns an ordered list of paths. `rete.yaml` is always first.
+pub fn discover_files(
```
wsnark, line 18: What about unit-testing this file discovery logic here?

src/bin/retectl.rs
wsnark: Well, current design seems quite good. Though I'd tune it towards `cargo build` behaviour:
- each error has its own `error` entry (with context -- but that should be done in a separate issue)
- last line summarizes, like `error: validation failed due to 3 previous parse error(s)`

Image

src/config/rete/model/fragment.rs
```diff
@@ -0,0 +1,28 @@
...
+/// file (e.g. two `services.api` entries) is a parse-time error, never a
+/// silent last-win.
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ConfigFragment {
+    pub source: Option<Source>,
+    pub rete: Option<Rete>,
+    pub nodes: Option<UniqueMap<String, Node>>,
+    pub services: Option<UniqueMap<String, Service>>,
+    /// Values are `Option<Group>` because reserved groups like `config-read:`
```
wsnark: I think this docstring is misleading: currently all groups have empty body, not only reserved ones. Need to fix all mentions. 

Also, is it possible to model as `<String, Group>` still, without Option, so that Group isn't different from other entities, just happens to be empty now?

src/config/rete/merge.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    let Some((_, rete)) = rete_block else {
+        let msg = if discovery_mode {
+            "No `rete` block found; rete.yaml must contain a `rete:` entry".into()
+        } else {
+            "No `rete` block found in the specified files; at least one must contain a `rete:` entry".into()
+        };
+        merge_errors.push(msg);
+        return Err(Report::new(LoadError::Merge(merge_errors.join("; "))));
```
wsnark: Maybe report them each on one line? Currently multiple entries look quite bad. Also shouldn't we list both offending files, not the second one only?

Image

src/config/rete/merge.rs
```diff
@@ -0,0 +1,342 @@
...
+
+use super::loader::LoadError;
+use super::model::{ConfigFragment, Group, Node, Rete, Role, Service, User};
+
+/// The merged whole-view of a rete config repository.
+///
+/// All collections are unioned from every source file; the `rete` block is
+/// the single canonical rete metadata entry.
+#[derive(Debug)]
+pub struct RepoModel {
```
wsnark: Shouldn't we put this within `model`?

src/config/rete/model/service.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Service {
+    /// Node this service runs on.
+    pub at: String,
+    /// Vertex on that node (optional when node has exactly one vertex).
+    pub via: Option<String>,
+    /// Local address the service binds to; flor forwards here.
+    pub addr: String,
```
wsnark: Maybe we should parse addresses to some struct, not just String? We could have validated "{DNS/IP}:port" early

src/config/rete/validate.rs
```diff
@@ -0,0 +1,1382 @@
...
+
+/// A single validation rule violation found in the rete config.
+pub struct Violation {
+    pub rule: Rule,
+    pub message: String,
+}
+
+/// The validation rule that was violated.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Rule {
```
wsnark, line 28: Actually we can have a numbered list of these Rules - like numbered list of errors `rustc` has, for example. But we need some separate doc maintaining such a list (and numbers should only increase; some may become deprecated...). Maybe create an issue for this task?

wsnark: Or we can model a linter here, instead. They don't have numbers, but a page with detailed descriptions is useful `https://rust-lang.github.io/rust-clippy/master/`...

src/config/rete/loader/discover.rs
wsnark: Though I see it is mostly covered by integration test, right?

wsnark (changes requested):

src/config/rete/source.rs
leviathan941: Discovery is a part of the `load` stage. Now `LoadOpts` is used only here. But you're right, it might be extended in the future, so it's better to be in the load module.

src/config/rete/source.rs
leviathan941: It refers to the Phase 1 in loader::load

src/config/rete/loader.rs
leviathan941: Discovery, merge, parse are stages of the load process, so yes, merge and source modules can be sub-modules of the loader module. I don't think this is necessary, but if you insist, I'll move it.

src/config/rete/model/service.rs
leviathan941: Do you propose to validate all config field type or just this one?

src/cli.rs
leviathan941: Is `print_error` OK name?

src/config/rete/validate.rs
leviathan941: We already have documentation with these rules:
https://florete.tech/docs/implementation/c0-tended-tunnels/mgmt-plane/validate-and-compile/#validator-rules
Do think it's not enough?

src/bin/retectl.rs
leviathan941: You're right. Need to think over it.

src/config/rete/source.rs
wsnark: I understand this, but such comments may become outdated, as there is no way to enforce such contract (Phase 1 in loader may become something different...). Just a nit, we can avoid this by not stating such relationships in the comments

src/config/rete/model/service.rs
wsnark: Everything that we can, I think, that is context-free (the names we validate explicitly). But AFAIK currently this is only address type; it shouldn't be hard - just parse into SockAddr?

src/config/rete/validate.rs
wsnark: I think we'd need a user-facing list of rules; so when a user sees "a rule A failed", we can point him for details to that item in a list. We shouldn't point him to design docs for sure... I have created a rete-labs/florete#111 to track this work

src/cli.rs
wsnark: `print_error` already exists, btw; Maybe we can merge them into single function that will do both logging and printing the error? Name isn't ideal, but acceptable probably. What I'd write (pseudo-code):
```rs
pub fn print_error<E>(report: &Report<E>) {
    log::error!("{report:?}");
    let style = error_prefix_style();
    if level >= debug { // determine from current log level, not from `--verbose`
        eprintln!("{style}error:{style:#} {report:?}");
    } else {
        eprintln!("{style}error:{style:#} {}", CompactChain(report));
    }
}
```

src/config/rete/loader.rs
wsnark: Let's move discover, parse and merge to submodules of the loader; and keep single LoadError here - I don't mind then. Currently it looks unclear, because validate is also at the same level as others, but actually it is not a submodule, and that's why validate error isn't listed here - and that's fine

src/bin/retectl.rs
leviathan941: My mistake was to connect writing error stack to the verbose command line key. I think this thing should be available for dev build only. I'll fix it.

src/bin/retectl.rs
leviathan941: Create new issue: https://github.com/rete-labs/flor/issues/52

src/config/rete/source.rs
leviathan941: I think you're right, let's use to_str instead

src/config/rete/model/fragment.rs
leviathan941: Option<T> is used by `serde` for deserialization of empty YAML body.
But the doc is misleading, you're right, because all groups are empty now.

src/config/rete/loader/discover.rs
leviathan941: Yes, discovery is covered by integration tests

src/bin/retectl.rs
leviathan941: Now logger will be initialized and a call stack printed for debug build only

src/config/rete/merge.rs
leviathan941: Fixed

src/config/rete/model/node.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub name: String,
+    pub kind: VertexKind,
+    #[serde(rename = "type")]
+    pub vertex_type: VertexType,
+    /// Listen address; absent on initiator-only nodes.
+    pub address: Option<SocketAddr>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
leviathan941: May be I need to use "snake_case" here?

wsnark: Agree, good catch!

wsnark (approved):
LGTM, nice

PR #41, opened 2026-06-25
feat: parse vertex json config (#39)

### LINKED ISSUE #39: Implement JSON config parser for `flor vertex`

wsnark:
Closes #39

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,5 @@ x509-parser = { version = "0.18", features = ["verify"] }
 assert_cmd = "2"
 ctor = "0.8"
 mockall = "0.14"
-serde_json = "1"
 serial_test = "3.4"
 tempfile = "3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,5 +37,6 @@ x509-parser = { version = "0.18", features = ["verify"] }
 assert_cmd = "2"
 ctor = "0.8"
 mockall = "0.14"
+serde_json = "1"
 serial_test = "3.4"
 tempfile = "3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ husky-rs = "0.3"
 log = { version = "0.4", features = ["std"] }
 mockall_double = "0.3"
 serde = { version = "1", features = ["derive"] }
+serde_json = "1"
 serde_yaml_ng = "0.10"
 thiserror = "2.0"
 tokio = { version = "1", features = ["full"] }
--- a/src/config/artifact/model/vertex.rs
+++ b/src/config/artifact/model/vertex.rs
@@ -62,6 +62,10 @@ impl Payload for VertexMgmtPayload {
 
 /// The local transport a vertex terminates. Internally tagged on `type`;
 /// unknown types fail closed.
+///
+/// Kept tagged (`{ "type": "quic" }`, not a bare `"quic"` string) so a transport
+/// can grow config — e.g. `{ "type": "quic", "idle_timeout": … }` — without a
+/// wire-format change, and so every variant-bearing node stays uniformly tagged.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
 pub enum TransportEndpoint {
@@ -157,8 +161,9 @@ pub enum Direction {
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
 pub enum LinkRule {
-    /// An explicit enumeration of links.
-    Enum { members: Vec<LinkMember> },
+    /// An explicit list of links — the members are enumerated directly (as
+    /// opposed to a future pattern/rule-matched set).
+    List { members: Vec<LinkMember> },
 }
 
 /// A single link: a named handle to a peer over one adapter.
@@ -232,7 +237,7 @@ mod tests {
                 }
             ],
             "links": [
-                { "type": "enum", "members": [
+                { "type": "list", "members": [
                     { "name": "config-server", "peer": "spiffe://rete-lovers/service/config-server", "via": { "type": "udp", "adapter": "wire", "addr": "9.10.11.12:4433" } },
                     { "name": "api",           "peer": "spiffe://rete-lovers/service/api",           "via": { "type": "udp", "adapter": "wire", "addr": "1.2.3.4:4433" } },
                     { "name": "kafka",         "peer": "spiffe://rete-lovers/service/kafka",         "via": { "type": "udp", "adapter": "wire", "addr": "5.6.7.8:4433" } }
@@ -278,7 +283,7 @@ mod tests {
                 { "target": "spiffe://rete-lovers/service/alpha/ssh", "allow": ["spiffe://rete-lovers/user/bob"] }
             ],
             "links": [
-                { "type": "enum", "members": [
+                { "type": "list", "members": [
                     { "name": "config-server", "peer": "spiffe://rete-lovers/service/config-server", "via": { "type": "udp", "adapter": "wire", "addr": "9.10.11.12:4433" } },
                     { "name": "mongodb",       "peer": "spiffe://rete-lovers/service/mongodb",       "via": { "type": "udp", "adapter": "wire", "addr": "5.6.7.8:4433" } }
                 ] }
@@ -324,7 +329,7 @@ mod tests {
 
         // One enum link rule with three members.
         assert_eq!(p.links.len(), 1);
-        let LinkRule::Enum { members } = &p.links[0];
+        let LinkRule::List { members } = &p.links[0];
         assert_eq!(members.len(), 3);
         assert_eq!(members[0].name, "config-server");
         assert_eq!(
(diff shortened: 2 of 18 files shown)


src/config/artifact/model/vertex.rs
```diff
@@ -0,0 +1,517 @@
...
+    pub io: Vec<IoChannel>,
+}
+
+/// A workload's on-disk identity material (scope-relative paths).
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Identity {
+    pub cert_path: PathBuf,
+    pub priv_path: PathBuf,
+}
```
leviathan941, line 124: Do we really pass a host local file system path in our config?

This is strange. How can an operator be sure where those things stored?

src/config/artifact/model/vertex.rs
```diff
@@ -0,0 +1,517 @@
...
+
+impl Payload for VertexMgmtPayload {
+    const PLANE: PlaneTag = PlaneTag::Mgmt;
+    const KIND: ArtifactKind = ArtifactKind::Vertex;
+}
+
+/// The local transport a vertex terminates. Internally tagged on `type`;
+/// unknown types fail closed.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
```
leviathan941, line 70: Why do you use internal tagging here? Is this done intentionally?

We get this:
```json
"transport_endpoint": { "type": "quic" }
```

But we able to get that:
```json
"transport_endpoint": "quic"
```
if we use external tagging instead.

src/config/artifact/validate.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+}
+
+/// Envelope-level checks, generic over any [`Payload`]: supported schema
+/// version, the `(plane, kind)` the payload declares, and the expected name.
+fn validate_envelope<P: Payload>(
+    env: &Envelope<P>,
+    expected_name: &str,
+) -> Result<(), Report<Error>> {
+    if env.schema_version != "1.0" {
```
leviathan941: Very strange check for the hard-coded version.

What is the expected evolution of this check?

src/config/artifact/validate.rs
```diff
@@ -0,0 +1,443 @@
...
+        bail!(Error::new(format!(
```
leviathan941: Don't we want to check everything possible and return all found errors?
`retectl validate` works that way.

src/config/artifact/model/vertex.rs
```diff
@@ -0,0 +1,517 @@
...
+    Bidirectional,
+}
+
+/// A `links` rule. Internally tagged on `type`; unknown types fail closed
+/// (no `#[serde(other)]`), so an unrecognised rule is an error, not a silent skip.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
+pub enum LinkRule {
+    /// An explicit enumeration of links.
+    Enum { members: Vec<LinkMember> },
```
leviathan941: `Enum` is not expected name for that collection I think.
May be `Array` or `List` are better? Wdyt?

Cargo.toml
```diff
@@ -33,5 +35,6 @@ x509-parser = { version = "0.18", features = ["verify"] }
 assert_cmd = "2"
 ctor = "0.8"
 mockall = "0.14"
+serde_json = "1"
```
leviathan941: We don't have to add a dependency in `dev-dependencies` if it's already in `dependencies`.

src/vertex.rs
```diff
@@ -0,0 +1,688 @@
...
+    let mut tcp = Vec::new();
+    for workload in &payload.workloads {
+        let svid = load_workload_svid(root, &workload.identity, &workload.spiffe_id)?;
+        for io in &workload.io {
+            match io {
+                IoChannel::Socks5 { listen } => socks5.push((svid.clone(), *listen)),
+                IoChannel::Tcp { upstream } => tcp.push((svid.clone(), *upstream)),
+                // FlorIO is the recursive bidirectional channel — it has no
+                // place in the inbound/outbound runtime, so it is not wired.
+                IoChannel::Florio { .. } => {}
```
leviathan941: We may panic here or at least write error log, don't we?

leviathan941 (changes requested):

src/config/artifact/model/vertex.rs
wsnark: Let's keep it internally tagged so we can easily add fields to it later, without schema-breaking changes

src/config/artifact/model/vertex.rs
wsnark: Renamed to `List`

src/vertex.rs
wsnark: Added a warning - I think it isn't a hard error...

src/config/artifact/model/vertex.rs
wsnark: Very good catch! That's a gap in our design. I think proper solution should be following:
- `coordinator` produces OS-agnostic artifact for `flor agent`
- `flor agent` produces OS-specific config for `flor vertex` it manages
Currently we're totally missing `flor agent` design. Created rete-labs/florete#116 to track this

src/config/artifact/validate.rs
wsnark: Very good catch. Suggested evolution:
1. Major version mismatch: error
2. Newer flor (e.g. 1.1) should accept older compatible schema  (1.0)
3. Older flor (e.g. 1.0) should reject newer schema (1.1) with meaningful message (because new schema may impose new bounds; fail-closed is a security property we want to maintain)

src/config/artifact/validate.rs
wsnark: Let's make this an enhancement. I feel like cost of rewriting into error collecting isn't justified at this point (the `rete validate` is what operator uses, not editing this JSON manually)

src/config/artifact/model/vertex.rs
leviathan941: So, keep this thing as is for now?

src/config/artifact/validate.rs
leviathan941: Ok, please create an issue then.

src/config/artifact/model/vertex.rs
wsnark: Yes, let's keep it as is, fixing when `flor agent` lands (it will need to translate configs anyway)

src/config/artifact/validate.rs
wsnark: Created #53

src/config/artifact/validate.rs
wsnark: Captured this generally in [ADR-0012](https://florete.tech/docs/implementation/adr/0012-version-artifact-contract/), implemented in code

src/vertex.rs
```diff
@@ -0,0 +1,704 @@
...
+    /// cannot skip validation or feed it the wrong kind.
+    pub fn load(root: &Path, name: &str) -> Result<ConfigBundle, Report<Error>> {
+        let path = root.join("mgmt/vertices").join(format!("{name}.json"));
+        let bytes = std::fs::read(&path).change_context_lazy(|| {
+            Error::new(format!("Failed to read vertex config {}", path.display()))
+        })?;
+        // Gate on schema_version before the strict parse, so a newer additive
+        // minor reports a clear upgrade error rather than an unknown-field
+        // failure from the envelope's `deny_unknown_fields`.
+        version::precheck_schema_version(&bytes).change_context_lazy(|| {
```
leviathan941: Could we reverse the order here and use `precheck_schema_version` only as a fallback when the strict typed parse fails?

The precheck is diagnostic rather than authoritative: its purpose is to replace an unknown field/unknown variant error with an actionable upgrade message when the artifact comes from a newer schema. Running it before the typed parse makes every valid artifact go through two full JSON passes.

Suggested flow:
1. Attempt the strict `Envelope` parse.
2. If it succeeds, let `Envelope::validate` perform the authoritative version check.
3. If it fails, run `precheck_schema_version`.
4. Return the unsupported-version error if detected; otherwise preserve the original parse error.

This keeps the intended diagnostics for newer schemas while avoiding the extra JSON pass on the successful path.

src/config/artifact/version.rs
```diff
@@ -0,0 +1,184 @@
...
+            let err = ensure_supported(v).unwrap_err();
+            assert!(
+                format!("{err:?}").contains("Malformed schema_version"),
+                "{v:?}: {err:?}"
+            );
+        }
+    }
+
+    #[test]
+    fn probe_gates_bytes_by_version() {
```
leviathan941, line 168: Could we also add an end-to-end regression test for the behavior this precheck is intended to provide?

The current test calls `precheck_schema_version` directly on a minimal object and only asserts that it returns an error. It does not verify the important integration invariant: an artifact with a newer minor version and a genuinely unknown envelope or payload field should be reported as an actionable “upgrade flor” error rather than as a Serde `unknown field` error.

Ideally, the test should exercise the actual loading/parsing path. It would also be useful to verify the complementary case: an unknown field under the currently supported version must preserve the original strict parse error, since that indicates a mis-stamped or malformed artifact rather than version skew.

leviathan941 (changes requested):

src/config/artifact/version.rs
wsnark: Implemented

src/vertex.rs
wsnark: Good call, implemented

leviathan941 (approved):
LGTM

Issue #43, opened 2026-06-27
Provide flor with rete trust domain source in JSON


wsnark:
Currently flor vertex's rete trust domain is inferred from its mgmt artifact's SPIFFE IDs (validated to be internally consistent, but no more). It should instead be sourced from rete-root metadata written by enrollment/flor agent (e.g. retes/<scope>/rete.json carrying trust_domain), read by both agent and vertex — so we can verify the IDs belong to our rete, not merely agree with each other. 

Until this is implemented, a wrong td fails closed at mTLS. 

Should be implemented with/after `flor agent` (#42).

wsnark:
Seems like this issue is implemented within #31's https://github.com/rete-labs/flor/pull/57/changes#diff-2b0dcf1999776820ae730e49ea165d68ef1d16ef6c3f538e9f6c214fa85d959c 

Need to confirm that old "derive trust domain from first SPIFFE ID" is gone, and close as done together with the compiler?

Issue #42, opened 2026-06-27
Implement minimal `flor agent`


wsnark:
- Implementation should be based on design per rete-labs/florete#116

wsnark:
The agent owns log-sink provisioning, per [ADR-0014](https://florete.tech/docs/implementation/adr/0014-provision-log-sinks) and [C0 Observability § The log stream](https://florete.tech/docs/implementation/c0-tended-tunnels/observability#the-log-stream) — it should be in scope here rather than a separate issue, since it happens at spawn.

Before fork, the agent opens a sink for each workload and installs it as fd 2:

- a `SOCK_DGRAM` connected to `/dev/log` where one exists;
- a per-workload `<workload>.log` append fd otherwise;
- its own fd 2 passed straight down when the agent was itself started in a terminal or with output redirected.

It then states the class in `FLOR_LOG_SINK` — the consuming half is #73. The agent is never in the data path and never sees a log byte.

PR #57, opened 2026-07-20
feat: implement retectl compile (#31)

### LINKED ISSUE #31: Implement minimal `retectl compile`
Compile step is described in https://florete.tech/docs/implementation/c0-tended-tunnels/mgmt-plane/validate-and-compile/#compile-step

It is OK to have some minimal version working, with further improvements split-off as new tickets

leviathan941:
Closes #31

### DIFF
--- a/src/bin/retectl.rs
+++ b/src/bin/retectl.rs
@@ -8,12 +8,14 @@
 use std::path::PathBuf;
 use std::time::Duration;
 
+use chrono::{SecondsFormat, Utc};
 use clap::{Args, Parser, Subcommand};
 use error_stack::{Report, ResultExt};
 
 use flor::{
     cli::{compact_chain, print_error, print_error_lines, write_secret},
-    config::rete::{LoadOpts, load, validate},
+    config::compile::{CompileOpts, compile, layout, version},
+    config::rete::{LoadOpts, RepoModel, load, validate},
     core::identity::{Ca, Kind, TrustDomain, build_id},
 };
 
@@ -37,6 +39,8 @@ enum Cmd {
     },
     /// Validate rete config: schema, cross-references, access consistency.
     Validate(ValidateArgs),
+    /// Compile rete config into per-node artifacts.
+    Compile(CompileArgs),
 }
 
 #[derive(Args, Debug)]
@@ -48,6 +52,21 @@ struct ValidateArgs {
     /// Override file discovery with explicit paths or globs (repeatable).
     #[arg(short, long = "file", value_name = "FILE")]
     files: Vec<String>,
+}
+
+#[derive(Args, Debug)]
+struct CompileArgs {
+    /// Repository root directory.
+    #[arg(long, default_value = ".")]
+    repo: PathBuf,
+
+    /// Override file discovery with explicit paths or globs (repeatable).
+    #[arg(short, long = "file", value_name = "FILE")]
+    files: Vec<String>,
+
+    /// Where to write the compiled tree (default: `<repo>/.flor/compiled`).
+    #[arg(long)]
+    out: Option<PathBuf>,
 }
 
 #[derive(Subcommand, Debug)]
@@ -123,55 +142,111 @@ fn main() {
     }
 
     if let Err(e) = run(cli.cmd) {
-        print_error(&e, DEBUG);
+        match e {
+            CmdError::Single(report) => print_error(&report, DEBUG),
+            CmdError::Many { stage, errors } => print_error_lines(&errors, stage),
+        }
         std::process::exit(1);
     }
 }
 
-fn run(cmd: Cmd) -> Result<(), Report<Error>> {
+/// How a command failed — one error, or a collected set of them.
+enum CmdError {
+    /// A single failure, reported as its causal chain.
+    Single(Report<Error>),
+    /// A stage that runs to completion and collects every failure it finds
+    /// (config load, validation), reported as a list. The `stage` names it in
+    /// the summary line; the errors are already rendered.
+    Many {
+        stage: &'static str,
+        errors: Vec<String>,
+    },
+}
+
+impl From<Report<Error>> for CmdError {
+    fn from(report: Report<Error>) -> Self {
+        CmdError::Single(report)
+    }
+}
+
+fn run(cmd: Cmd) -> Result<(), CmdError> {
     match cmd {
         Cmd::Ca { action } => match action {
             CaAction::Init(args) => ca_init(args),
             CaAction::Sign(args) => ca_sign(args),
         },
         Cmd::Validate(args) => cmd_validate(args),
+        Cmd::Compile(args) => cmd_compile(args),
     }
 }
 
-fn cmd_validate(args: ValidateArgs) -> Result<(), Report<Error>> {
-    let opts = LoadOpts {
-        repo: args.repo,
-        files: args.files,
+fn cmd_validate(args: ValidateArgs) -> Result<(), CmdError> {
+    load_valid_model(args.repo, args.files)?;
+    println!("ok");
+    Ok(())
+}
+
+/// `compile` writes artifacts an agent will act on, so it refuses anything
+/// `validate` would reject: same load, same rules, then the projection.
+fn cmd_compile(args: CompileArgs) -> Result<(), CmdError> {
+    let out = args
+        .out
+        .unwrap_or_else(|| args.repo.join(".flor").join("compiled"));
+    let model = load_valid_model(args.repo, args.files)?;
+
+    // Read the counter off the tree before the write clears it.
+    let opts = CompileOpts {
+        version: version::next_version(&out)
+            .change_context(Error("Failed to read the compiled version".into()))?,
+        generated_at: Utc::now().to_rfc3339_opts(SecondsFormat::Secs, true),
     };
 
-    let model = match load(&opts) {
-        Ok(model) => model,
-        Err(failures) => {
-            for f in &failures {
-                log::error!("{f:?}");
-            }
-            let lines: Vec<String> = failures.iter().map(compact_chain).collect();
-            print_error_lines(&lines, "load");
-            std::process::exit(1);
+    let artifacts =
+        compile(&model, &opts).change_context(Error("Failed to compile rete config".into()))?;
+    let written = layout::write(&out, &artifacts)
+        .change_context(Error("Failed to write the compiled tree".into()))?;
+
+    for path in &written {
+        println!("{}", path.display());
+    }
+    println!(
+        "Compiled {} node(s) at version {}",
+        artifacts.len(),
+        opts.version
+    );
+    Ok(())
+}
+
+/// Load the merged source and run the validator over it. The `Ok` model is one
+/// every rule accepted — the guarantee `compile` builds on.
+fn load_valid_model(repo: PathBuf, files: Vec<String>) -> Result<RepoModel, CmdError> {
+    let opts = LoadOpts { repo, files };
+
+    let model = load(&opts).map_err(|failures| {
+        for f in &failures {
+            log::error!("{f:?}");
         }
-    };
+        CmdError::Many {
+            stage: "load",
+            errors: failures.iter().map(compact_chain).collect(),
+        }
+    })?;
 
     let violations = validate(&model);
-
-    if violations.is_empty() {
-        println!("ok");
-        return Ok(());
+    if !violations.is_empty() {
+        return Err(CmdError::Many {
+            stage: "validation",
+            errors: violations
+                .iter()
+                .map(|v| format!("[{}] {}", v.rule, v.message))
+                .collect(),
+        });
     }
 
-    let lines: Vec<String> = violations
-        .iter()
-        .map(|v| format!("[{}] {}", v.rule, v.message))
-        .collect();
-    print_error_lines(&lines, "validation");
-    std::process::exit(1);
+    Ok(model)
 }
 
-fn ca_init(args: CaInitArgs) -> Result<(), Report<Error>> {
+fn ca_init(args: CaInitArgs) -> Result<(), CmdError> {
     let td = TrustDomain::new(&args.trust_domain)
         .change_context(Error("Invalid trust domain".into()))?;
     let ca =
@@ -188,7 +263,7 @@ fn ca_init(args: CaInitArgs) -> Result<(), Report<Error>> {
     Ok(())
 }
 
-fn ca_sign(args: CaSignArgs) -> Result<(), Report<Error>> {
+fn ca_sign(args: CaSignArgs) -> Result<(), CmdError> {
     let cert_pem_bytes = std::fs::read(&args.ca_cert)
         .change_context_lazy(|| Error(format!("Failed to read {}", args.ca_cert.display())))?;
     let key_pem_bytes = std::fs::read(&args.ca_key)
--- a/src/config/artifact/model/vertex.rs
+++ b/src/config/artifact/model/vertex.rs
@@ -12,13 +12,14 @@
 //! `kind`. An io channel's direction is intrinsic to its `kind` (no
 //! `direction` field) — see [`IoChannel::direction`].
 //!
-//! **No identity paths.** A workload names its principal by SPIFFE ID and the
-//! node's [identity store](crate::core::identity::Store) materializes it, so the
-//! same compiled JSON is valid wherever the scope root lives and on any OS. The
-//! only paths here are local socket addresses a FlorIO channel binds.
+//! **No filesystem paths at all.** A workload names its principal by SPIFFE ID
+//! and the node's [identity store](crate::core::identity::Store) materializes
+//! it; a FlorIO adapter names the *vertex* that serves it, and the socket path
+//! is a runtime convention over the spawn-injected runtime root
+//! (`<runtime-root>/vertices/<vertex-name>.sock`), never a compiled field. So
+//! the same compiled JSON is valid wherever the scope root lives and on any OS.
 
 use std::net::SocketAddr;
-use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
 
@@ -103,8 +104,10 @@ pub enum Adapter {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         listen: Option<SocketAddr>,
     },
-    /// A FlorIO socket that delegates dialing to a layer below.
-    Florio { name: String, socket: PathBuf },
+    /// A FlorIO channel that delegates dialing to the layer below, naming the
+    /// vertex that serves it — the socket path is derived at runtime from that
+    /// name and the injected runtime root, never carried here.
+    Florio { name: String, vertex: String },
 }
 
 impl Adapter {
@@ -140,8 +143,9 @@ pub enum IoChannel {
     Socks5 { listen: SocketAddr },
     /// Outbound: flor delivers inbound connections to this upstream.
     Tcp { upstream: SocketAddr },
-    /// Bidirectional: a recursive flor layer.
-    Florio { socket: PathBuf },
+    /// Bidirectional: a recursive flor layer. Carries nothing — the serving
+    /// vertex derives the socket path from its own name at runtime.
+    Florio {},
 }
 
 impl IoChannel {
@@ -425,15 +429,15 @@ mod tests {
 
     #[test]
     fn florio_adapter_and_io_parse() {
-        // Exercise the FlorIO variants (socket-bearing, bidirectional).
+        // Exercise the FlorIO variants (vertex-named, bidirectional).
         let v = json!({
             "schema_version": "1.0",
             "kind": "mesh",
             "transport_endpoint": { "type": "quic" },
-            "connection_manager": { "adapters": [ { "name": "io", "type": "florio", "socket": "/run/flor.sock" } ] },
+            "connection_manager": { "adapters": [ { "name": "io", "type": "florio", "vertex": "public" } ] },
             "workloads": [
                 { "spiffe_id": "spiffe://rete-lovers/vertex/alpha/rete",
-                  "io": [ { "kind": "florio", "socket": "/run/wl.sock" } ] }
+                  "io": [ { "kind": "florio" } ] }
             ],
             "links": [
                 { "type": "list", "members": [
@@ -443,9 +447,9 @@ mod tests {
         });
         let p: VertexMgmtPayload = serde_json::from_value(v).unwrap();
         match &p.connection_manager.adapters[0] {
-            Adapter::Florio { name, socket } => {
+            Adapter::Florio { name, vertex } => {
                 assert_eq!(name, "io");
-                assert_eq!(socket, &PathBuf::from("/run/flor.sock"));
+                assert_eq!(vertex, "public");
             }
             other => panic!("expected florio adapter, got {other:?}"),
         }
@@ -478,10 +482,11 @@ mod tests {
     }
 
     #[test]
-    fn rejects_the_dropped_identity_path_fields() {
-        // Artifacts carry no filesystem references. The closed schema is what
-        // stops an artifact compiled against the older shape from loading with
-        // its paths silently ignored.
+    fn rejects_the_dropped_path_fields() {
+        // Artifacts carry no filesystem references — identity material and
+        // FlorIO sockets alike. The closed schema is what stops an artifact
+        // compiled against the older shape from loading with its paths
+        // silently ignored.
         let mut v = user_node();
         v["ca_cert_path"] = json!("ca.crt");
         assert!(
@@ -494,6 +499,21 @@ mod tests {
         assert!(
             serde_json::from_value::<VertexMgmtPayload>(v).is_err(),
             "per-workload identity paths must be rejected"
+        );
+
+        let mut v = user_node();
+        v["connection_manager"]["adapters"] =
+            json!([ { "name": "io", "type": "florio", "socket": "/run/flor.sock" } ]);
+        assert!(
+            serde_json::from_value::<VertexMgmtPayload>(v).is_err(),
+            "a florio adapter socket path must be rejected"
+        );
+
+        let mut v = user_node();
+        v["workloads"][0]["io"] = json!([ { "kind": "florio", "socket": "/run/wl.sock" } ]);
+        assert!(
+            serde_json::from_value::<VertexMgmtPayload>(v).is_err(),
+            "a florio io socket path must be rejected"
         );
     }
 
--- a/src/config/artifact/model/vertex.rs
+++ b/src/config/artifact/model/vertex.rs
@@ -19,6 +19,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::core::identity::SpiffeId;
 
+use super::super::version::{self, Contract};
 use super::envelope::{ArtifactKind, Payload, PlaneTag};
 
 /// Which forwarding engine a vertex runs. A plain discriminator: the mgmt
@@ -34,6 +35,10 @@ pub enum VertexKind {
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct VertexMgmtPayload {
+    /// `major.minor` of the vertex payload-family contract this content uses —
+    /// the family's own ladder, independent of the envelope's. Declared first so
+    /// it serializes first, as the C0 artifact examples show it.
+    pub schema_version: String,
     /// Which engine this vertex runs.
     pub kind: VertexKind,
     /// Scope-relative path to the rete CA certificate.
@@ -58,6 +63,11 @@ pub struct VertexMgmtPayload {
 impl Payload for VertexMgmtPayload {
     const PLANE: PlaneTag = PlaneTag::Mgmt;
     const KIND: ArtifactKind = ArtifactKind::Vertex;
+    const FAMILY: Contract = version::VERTEX;
+
+    fn schema_version(&self) -> &str {
+        &self.schema_version
+    }
 }
 
 /// The local transport a vertex terminates. Internally tagged on `type`;
@@ -220,6 +230,7 @@ mod tests {
     /// The exact C0 user-node link payload from validate-and-compile.mdx.
     fn user_node() -> Value {
         json!({
+            "schema_version": "1.0",
             "kind": "link",
             "ca_cert_path": "ca.crt",
             "transport_endpoint": { "type": "quic" },
@@ -254,6 +265,7 @@ mod tests {
     /// The exact C0 server-node link payload from validate-and-compile.mdx.
     fn server_node() -> Value {
         json!({
+            "schema_version": "1.0",
             "kind": "link",
             "ca_cert_path": "ca.crt",
             "transport_endpoint": { "type": "quic" },
@@ -406,6 +418,7 @@ mod tests {
     fn parses_mesh_payload() {
         // Same struct, `kind: mesh`.
         let v = json!({
+            "schema_version": "1.0",
             "kind": "mesh",
             "ca_cert_path": "ca.crt",
             "transport_endpoint": { "type": "quic" },
@@ -427,6 +440,7 @@ mod tests {
     fn florio_adapter_and_io_parse() {
         // Exercise the FlorIO variants (socket-bearing, bidirectional).
         let v = json!({
+            "schema_version": "1.0",
             "kind": "mesh",
             "ca_cert_path": "ca.crt",
             "transport_endpoint": { "type": "quic" },
--- a/src/config/artifact/model/vertex.rs
+++ b/src/config/artifact/model/vertex.rs
@@ -11,6 +11,11 @@
 //! internally tagged: adapters and link rules on `type`, io channels on
 //! `kind`. An io channel's direction is intrinsic to its `kind` (no
 //! `direction` field) — see [`IoChannel::direction`].
+//!
+//! **No identity paths.** A workload names its principal by SPIFFE ID and the
+//! node's [identity store](crate::core::identity::Store) materializes it, so the
+//! same compiled JSON is valid wherever the scope root lives and on any OS. The
+//! only paths here are local socket addresses a FlorIO channel binds.
 
 use std::net::SocketAddr;
 use std::path::PathBuf;
@@ -20,7 +25,7 @@ use serde::{Deserialize, Serialize};
 use crate::core::identity::SpiffeId;
 
 use super::super::version::{self, Contract};
-use super::envelope::{ArtifactKind, Payload, PlaneTag};
+use super::envelope::{Payload, PlaneTag};
 
 /// Which forwarding engine a vertex runs. A plain discriminator: the mgmt
 /// payload is field-identical across both.
@@ -41,8 +46,6 @@ pub struct VertexMgmtPayload {
     pub schema_version: String,
     /// Which engine this vertex runs.
     pub kind: VertexKind,
-    /// Scope-relative path to the rete CA certificate.
-    pub ca_cert_path: PathBuf,
     /// The local transport the vertex terminates.
     pub transport_endpoint: TransportEndpoint,
     /// Local transport mechanisms a link can dial over.
@@ -62,7 +65,6 @@ pub struct VertexMgmtPayload {
 
 impl Payload for VertexMgmtPayload {
     const PLANE: PlaneTag = PlaneTag::Mgmt;
-    const KIND: ArtifactKind = ArtifactKind::Vertex;
     const FAMILY: Contract = version::VERTEX;
 
     fn schema_version(&self) -> &str {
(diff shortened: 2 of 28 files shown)


src/bin/retectl.rs
```diff
@@ -123,55 +142,111 @@ fn main() {
...
+        }
         std::process::exit(1);
     }
 }
 
-fn run(cmd: Cmd) -> Result<(), Report<Error>> {
+/// How a command failed — one error, or a collected set of them.
+enum CmdError {
+    /// A single failure, reported as its causal chain.
+    Single(Report<Error>),
```
wsnark, line 156: What about using native error_stack's `Report<[C]>` for multiple errors? https://docs.rs/error-stack/latest/error_stack/#multiple-errors

If  this makes sense, let's create an enhancement request

src/bin/retectl.rs
```diff
@@ -123,55 +142,111 @@ fn main() {
...
+/// `validate` would reject: same load, same rules, then the projection.
+fn cmd_compile(args: CompileArgs) -> Result<(), CmdError> {
+    let out = args
+        .out
+        .unwrap_or_else(|| args.repo.join(".flor").join("compiled"));
+    let model = load_valid_model(args.repo, args.files)?;
+
+    // Read the counter off the tree before the write clears it.
+    let opts = CompileOpts {
+        version: version::next_version(&out)
```
wsnark, line 199: This needs to be upgraded to per-node version (worth a ticket, not here)

src/config/rete/model/user.rs
```diff
@@ -21,4 +23,6 @@ pub struct UserNode {
     pub at: String,
     /// Vertex on that node (optional when node has exactly one vertex).
     pub via: Option<String>,
+    /// Loopback SOCKS5 port flor exposes for this device's outbound calls.
```
wsnark, line 26: Create a ticket to fix design docs to include this feature (eventually it may be reshaped a bit, to allow easy extension with new inbounds - but we'll fix after this redesign then; it must go through both users and services - currently they're similar - and both lack this extensibility)

src/config/rete/validate.rs
```diff
@@ -528,6 +530,56 @@ fn check_principal_role_coherence(model: &RepoModel, out: &mut Vec<Violation>) {
     }
 }
 
+// ---------------------------------------------------------------------------
+// Rule::LocalPortUniqueness
```
wsnark: Need a ticket to update the design doc to include this validator rule, right?

src/config/rete/validate.rs
```diff
@@ -528,6 +530,56 @@ fn check_principal_role_coherence(model: &RepoModel, out: &mut Vec<Violation>) {
...
+    use std::net::SocketAddr;
+
+    // (node, listen) -> the subject that first claimed it.
+    let mut claimed: HashMap<(&str, SocketAddr), String> = HashMap::new();
+
+    let listeners = model
+        .services
+        .iter()
+        .filter_map(|(name, svc)| {
+            svc.socks5_proxy
```
wsnark: We need to check svc.addr too - outbounds share socket space with inbounds. 

BTW, maybe we want to be more strict and should not allow 0.0.0.0:port and 127.0.0.1:port (behaviour is OS- and bind order-specific, so rather undefined to us)? Then we need to include UDP vertex SockAddresses too, and check cross-IP ports... 

I suggest to file a ticket that will be focused on defining proper validation as it seems quite complicated: what other addrs are special? We have to support IPv6 too - and decide whether to allow a mix of them; this is OS and OS-config-specific, it seems! We need to research the topic and provide some common ground, to prevent subtle instabilities (maybe we can allow an operator's override if he wants to mix consciously - but that's a feature/enh request)...  

Also do we allow non-local inbounds? That seems to be strange, in C0/C1 it seems meaningless and should be prohibited (if ever we get non-local inbounds, this can be amended, right?...). But for outbounds that's perfectly fine, though maybe worth an explicit YAML config claim, so that normal services are local, and external ones are explicitly set as such by operator? Another topic for research.

src/config/rete/validate.rs
```diff
@@ -528,6 +530,56 @@ fn check_principal_role_coherence(model: &RepoModel, out: &mut Vec<Violation>) {
...
 }
 
+// ---------------------------------------------------------------------------
+// Rule::LocalPortUniqueness
+// ---------------------------------------------------------------------------
+
+fn check_local_port_uniqueness(model: &RepoModel, out: &mut Vec<Violation>) {
+    // Services and user devices each declare a loopback SOCKS5 listener. On a
+    // given node no two may share an address, or the second would fail to bind.
+    // (The node agent's listener is allocated by the compiler to avoid these,
```
wsnark: Probably we need a way to override this agent's inbound port too? A feature request (and later on we'd have a local port negotiation feature - so these compiler/operator-assigned local ports of inbounds become recommendations which can be re-bound by vertices if their addrs are in use)

src/config/rete/validate.rs
```diff
@@ -1381,4 +1436,90 @@ mod tests {
...
+
+    #[test]
+    fn local_port_uniqueness_user_and_service_share_a_port() {
+        let mut model = minimal_valid_model();
+        model.services.insert(
+            "client-svc".into(),
+            Service {
+                at: "mgmt".into(),
+                via: None,
+                addr: "127.0.0.1:8080".parse().unwrap(),
```
wsnark: This is a bad test: it has 3 non-unique addrs, but the check is across only 2 of them. Misleading

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+//! The resolved rete: everything the per-node projection needs, computed once.
+//!
+//! [`Plan::build`] does the work that is rete-wide rather than node-local —
+//! naming every principal and target with its SPIFFE ID, expanding roles to the
+//! groups they grant, resolving each service to the wire address that dials it,
+//! and allocating the local SOCKS5 ports the source never spells out. The
+//! per-node projection ([`super::vertex`]) then only filters and formats.
+//!
+//! Ordering is deliberate: the source model's collections are `HashMap`s, so
+//! every collection here is a `BTree*` or an explicitly sorted `Vec`. Same
```
wsnark: Need to rephrase to be more explicit: source model's collections are HashMaps that do not have deterministic order, but here we use BTree/sorted Vecs to ensure determinism

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+    pub signer_key_id: SpiffeId,
+    /// Every node, by name.
+    pub nodes: BTreeMap<String, NodePlan>,
+    /// Every service as a dialable target, by service name.
+    pub targets: BTreeMap<String, Target>,
+    /// Every principal in the rete, sorted by SPIFFE ID then by node.
+    ///
+    /// A multi-device user contributes one entry per device node: same identity,
+    /// different local io.
+    pub principals: Vec<Principal>,
```
wsnark: Principal is very broad thing: https://florete.tech/docs/implementation/c0-tended-tunnels/mgmt-plane/identity/#types-of-principal - is it used here properly? Very unclear.. Maybe these are TlsPrincipals? But then, they include targets too - so maybe these are Initiators (counterpart of Target principals)?

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+/// A node's link vertex — the one artifact C0 compiles per node.
```
wsnark: C0 compiles up to 3 artifacts per node. Need to file 2 tickets -- to implement compiler to emit agent's artifact + coordinator's artifact (for mgmt node only)

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+#[derive(Debug)]
+pub struct NodePlan {
+    /// The vertex's name in `nodes.yaml`; also the artifact's `name` and filename.
```
wsnark: nodes.yaml is vague - it may be absent... need to use better wording. The `name` is very important - it is actually a SpiffeId's name segment always now (link vertex has SpiffeID too, but it isn't TlsPrincipal, so it doesn't have SVID(cert+key))

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+impl NodePlan {
+    /// What the wire adapter binds: the declared port on an unspecified host.
+    ///
+    /// `nodes.yaml` states where peers *reach* the node; the public IP may not
+    /// exist on any local NIC (NAT, load balancer, floating address), so what
+    /// the socket binds is the compiler's business.
+    pub fn listen(&self) -> Option<SocketAddr> {
+        self.address.map(|address| {
+            let host = match address {
+                SocketAddr::V4(_) => IpAddr::V4(Ipv4Addr::UNSPECIFIED),
```
wsnark: This is really strange. I think we should support operator-authored listen addresses here: e.g. for a multi-interface node the operator selects at which interface this particular link listens - it mustn't listen on UNSPECIFIED/0.0.0.0 (unless the operator says so - he can easily do this explicitly)!

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+/// A service seen from the outside: who may reach it, and how to dial it.
+#[derive(Debug)]
+pub struct Target {
+    /// The service's name in `services.yaml`; also its local forwarding handle.
+    pub name: String,
+    pub id: SpiffeId,
+    /// The node hosting it.
+    pub node: String,
+    /// The host node's declared address — what an initiator's link dials.
+    pub addr: SocketAddr,
```
wsnark: This naming (`addr`) is misleading -- currently YAML service has `addr` - that is its upstream address. Here `addr` is QUIC link vertex''s `address`. I suggest renaming (and updating the docstring, as there is no "host node's declared address" term at all - that's QUIC link vertex's address).

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// Scope-relative cert and key, resolved by the agent against the rete root.
+    pub identity: Identity,
+    /// How the principal is wired into flor locally.
+    pub io: Vec<IoChannel>,
+}
+
+/// A service seen from the outside: who may reach it, and how to dial it.
+#[derive(Debug)]
+pub struct Target {
+    /// The service's name in `services.yaml`; also its local forwarding handle.
```
wsnark: don't refer to services.yaml either (may be absent, misleading)

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+                SocketAddr::V6(_) => IpAddr::V6(Ipv6Addr::UNSPECIFIED),
+            };
+            SocketAddr::new(host, address.port())
+        })
+    }
+}
+
+/// A principal as wired into one node's vertex.
+#[derive(Debug)]
+pub struct Principal {
```
wsnark: Looks like TlsPrincipal or Initiator only -- still unclear from the struct itself. `io` suggests that it includes both initiators and targets ... But separate Target confuses the reader.

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+/// A node's link vertex — the one artifact C0 compiles per node.
+#[derive(Debug)]
+pub struct NodePlan {
+    /// The vertex's name in `nodes.yaml`; also the artifact's `name` and filename.
+    pub vertex_name: String,
+    /// Where peers reach this node. `None` for initiator-only nodes (no
+    /// `address` — laptops behind NAT).
+    pub address: Option<SocketAddr>,
```
wsnark: This will not survive upgrade to non-IP nodes like radio ones in B2... probably OK for now, we'll have to refactor...

wsnark: But somehow the compiler reduces already existing abstraction of vertices that have addresses, not nodes. I feel like we should keep all abstractions that are defined in the design - and this will allow easier upgrades later, also avoid inconsistent understanding (code says "node address", while Florete says the node isn't part of the (TLS) communications and doesn't have an address...)

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+        )));
+    };
+    if links.next().is_some() {
+        bail!(Error::new(format!(
+            "Node '{name}' has more than one `kind: link, type: quic` vertex; \
+             C0 requires exactly one"
+        )));
+    }
+
+    Ok(NodePlan {
```
wsnark, line 219: Here is abstraction's collapse; let's keep vertices, do not reduce Node to vertex_name+address

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+    })
+}
+
+/// Every principal whose identity material lives on `node`: the users with a
+/// device here and the services hosted here.
+///
+/// Both name their own SOCKS5 port — services in `services.yaml`, users in each
+/// device entry under `users.yaml` — so nothing is allocated here; uniqueness of
+/// those ports is a validator concern ([`super::super::rete::validate`]).
+fn node_principals(
```
wsnark: Here as docstring speaks about SOCKS5 port, we're talking about Initiators, not broad Principals - right?

wsnark: Well, the code below takes both target and initiator services; unclear why targets are collected too - it depends on how the resulting array is used (targets-only will have empty `granted`)... So need to decide: are targets really needed in this function's output, or can be skipped; if needed, than it is TlsPrincipals, Initiators otherwise

wsnark: Ok, checked the calling code - these are used as TlsPrincipals, uniformly

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+    Ok(NodePlan {
+        vertex_name: vertex.name.clone(),
+        address: vertex.address,
+    })
+}
+
+/// Every principal whose identity material lives on `node`: the users with a
+/// device here and the services hosted here.
+///
+/// Both name their own SOCKS5 port — services in `services.yaml`, users in each
```
wsnark: This is misleading - services may have no SOCKS5 port. Better remove this note about socks altogether

src/config/compile/vertex.rs
```diff
@@ -0,0 +1,180 @@
+// Copyright (C) 2026 ReteLabs LLC.
+// Licensed under Apache-2.0 or MIT at your option.
+
+//! Projection of the resolved [`Plan`] onto one node's vertex artifact.
+//!
+//! Per-node filtering is the whole point: a node's artifact carries only the
+//! identity material and ACL rows its own workloads need. Three tables fall out
```
wsnark: node's artifact doesn't carry any identity material (that implies SVIDs to me - i.e. certs/keys); it carries only Spiffe IDs - need to say that explicitly, to avoid wrong implications

src/config/compile/vertex.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::config::artifact::{
+    ArtifactKind, Envelope, Plane, Signature, VertexKind, VertexMgmtPayload,
+};
+use crate::core::identity::SpiffeId;
+
+use super::plan::{NodePlan, Plan, Principal, Target};
+use super::{CompileOpts, Error, NodeVertexArtifact};
+
+/// The compiled-artifact contract version this compiler emits.
+const SCHEMA_VERSION: &str = "1.0";
```
wsnark: Need to file a ticket to fix schema version generation: Envelope will have it's own ENVELOPE_SCHEMA_VERSION; payloads will have their own - VERTEX_SCHEMA_VERSION, AGENT_SCHEMA_VERSION, COORDINATOR_SCHEMA_VERSION. 

Moreover, there is a requirement to emit least-needed schema version based on features in use (so that schema bump affects only those nodes which need these features, not the whole fleet... that may be an enh/feature request by itself for later).

src/config/compile/vertex.rs
```diff
@@ -0,0 +1,180 @@
...
+/// The rete CA, as the flat install root holds it.
+const CA_CERT_FILE: &str = "ca.crt";
+
+/// A link vertex has exactly one adapter — one QUIC endpoint over one UDP
+/// socket. Its name is a local handle each link's `via` references.
+const WIRE_ADAPTER: &str = "wire";
+
+/// Signing is not implemented yet: artifacts name the rete's mgmt signer but
+/// carry no signature. The agent — the sole verifier — lands with `flor agent`.
+const SIGNATURE_ALG_NONE: &str = "none";
```
wsnark, line 39: A ticket to implement signing is needed

src/config/compile.rs
```diff
@@ -0,0 +1,493 @@
...
+//! be written under `<repo>/.flor/compiled/<node>/mgmt/vertices/<vertex>.json`.
+//!
+//! Two stages, mirroring what the compiled artifact needs to say:
+//! - [`plan`] resolves the whole rete once — SPIFFE IDs, role→group expansion,
+//!   local port allocation, per-node placement.
+//! - [`vertex`] projects that resolved view onto a single node, keeping only the
+//!   identity material and ACL rows relevant to its own workloads.
+//!
+//! The projection is pure and deterministic: the source model's collections are
+//! `HashMap`s, so every list this module emits is sorted. The one module that
```
wsnark: This needs to be rephrased: `HashMap`s are NOT deterministic, so we explicitly sort ...

src/config/compile.rs
```diff
@@ -0,0 +1,493 @@
...
+#[derive(Debug, Clone)]
+pub struct NodeVertexArtifact {
+    pub node: String,
+    pub vertex_name: String,
+    pub envelope: Envelope<VertexMgmtPayload>,
+}
+
+impl NodeVertexArtifact {
+    /// Where this artifact lives, relative to the compiled tree root:
+    /// `<node>/mgmt/vertices/<vertex>.json`.
```
wsnark: the path now is flat, no "vertices" dir - probably need to fix right here

src/config/compile/vertex.rs
```diff
@@ -0,0 +1,180 @@
...
+            .filter(|p| grants_access(&p.granted, &target.groups))
+            .map(|p| &p.id)
+            .collect();
+        if !allow.is_empty() {
+            reachable.insert(&target.name, (target, allow));
+        }
+    }
+
+    let payload = VertexMgmtPayload {
+        kind: VertexKind::Link,
```
wsnark, line 61: Kind is dropped from the Envelope schema, need to fix (either here or in some follow-up ticket)

src/config/compile/vertex.rs
```diff
@@ -0,0 +1,180 @@
...
+        ca_cert_path: CA_CERT_FILE.into(),
```
wsnark: Same for all paths - they're removed

src/config/compile/vertex.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/// Sort and de-duplicate an allow-list. A user with several devices appears once
+/// per device in the plan, but names a single identity in an ACL.
+fn sorted_ids<'a>(ids: impl Iterator<Item = &'a SpiffeId>) -> Vec<SpiffeId> {
+    let mut out: Vec<SpiffeId> = ids.cloned().collect();
+    out.sort_by_key(|id| id.to_string());
+    out.dedup();
+    out
+}
```
wsnark, line 193: What about adding unit tests to cover all meaningful functions (ingress, egress, links, and plan itself - which maybe can split out one more function `reachable()` for testability)?

src/config/compile/plan.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+/// A principal's identity material: bare filenames, resolved by the agent
+/// against the (flat) rete install root.
+fn identity_of(name: &str) -> Identity {
+    Identity {
+        cert_path: format!("{name}.crt").into(),
+        priv_path: format!("{name}.key").into(),
+    }
+}
```
wsnark, line 454: What about unit-testing `build`, `granted_groups` and `node_principals`?

wsnark (changes requested):

src/config/rete/validate.rs
leviathan941: Not all written inside a test must be tested. A test should test only one thing. Sometimes you need to setup a complex environment to test that only thing.
I've added the comment to the code that this address is not tested here. It should be enough I guess.

src/config/compile/plan.rs
leviathan941: `NodePlan` here is not actually a node. Since, now only vertex is supported it's VertexPlan. It might be separated to NodePlan which consists of VertexPlan only.

src/bin/retectl.rs
leviathan941: I don't think that we win something there.

- `CmdError` is about `how-to-render`, not `how-to-pass`. So we will have to keep it anyway to choose how to write errors to user.
- `config::rete::validate::validate` function. Now it returns `Vec<Violation>`. Violation is about a rule, it has no source, no chain, etc. So, it's not an Error actually. We can make it Error or course, but I don't think we need this.
- `config::rete::loader::load` - the only thing where error_stack's multiple errors really make sense. But I don't think code will be simpler when we apply that mechanism.

So, I can create a ticket for switching `config::rete::loader::load` function to that mechanism. But I don't think it worth it.

wsnark: Agreed, let's keep this as is; error_stack feature may be useful elsewhere later

src/config/compile/plan.rs
wsnark: I'd approve changing it to VertexPlan, and wrapping in a NodePlan - that would be correct abstraction

src/bin/retectl.rs
leviathan941: https://github.com/rete-labs/flor/issues/59

src/config/compile/plan.rs
leviathan941: https://github.com/rete-labs/flor/issues/60
https://github.com/rete-labs/flor/issues/61

src/config/compile/vertex.rs
leviathan941: Done

src/config/compile/vertex.rs
leviathan941: https://github.com/rete-labs/flor/issues/62

src/config/rete/validate.rs
leviathan941: https://github.com/rete-labs/flor/issues/64

src/config/compile/plan.rs
leviathan941: Removed that strange solution and created new ticket to design a correct one: https://github.com/rete-labs/flor/issues/63

src/config/rete/validate.rs
leviathan941: Removed current partial solution and created new ticket: https://github.com/rete-labs/flor/issues/65

src/vertex.rs
```diff
@@ -62,14 +62,18 @@ pub struct ConfigBundle {
...
     /// Load and build the runtime config for the vertex named `name` under the
-    /// rete root `root`: read `mgmt/vertices/<name>.json`, validate it, and
-    /// (for a link vertex) build the bundle, loading identity material from
-    /// `root`. A mesh vertex errors (its runtime is not implemented yet).
+    /// scope root `root`: read `mgmt/<name>.json`, validate it, and (for a link
+    /// vertex) build the bundle, resolving identity through the scope's store.
+    /// A mesh vertex errors (its runtime is not implemented yet).
+    ///
+    /// The mgmt set is flat — one `<name>.json` per workload — which is where
+    /// `retectl compile` puts it (see
```
wsnark: It is wrong refer to `retectl compile` within `flor vertex` - the vertex process will receive artifacts from `flor agent` which places them flat, that's correct.

src/config/rete/validate.rs
leviathan941: https://github.com/rete-labs/flor/issues/65

src/config/artifact/model/vertex.rs
```diff
@@ -11,6 +11,11 @@
 //! internally tagged: adapters and link rules on `type`, io channels on
 //! `kind`. An io channel's direction is intrinsic to its `kind` (no
 //! `direction` field) — see [`IoChannel::direction`].
+//!
+//! **No identity paths.** A workload names its principal by SPIFFE ID and the
+//! node's [identity store](crate::core::identity::Store) materializes it, so the
+//! same compiled JSON is valid wherever the scope root lives and on any OS. The
+//! only paths here are local socket addresses a FlorIO channel binds.
```
wsnark: That local socket addresses (florio) are also removed (no paths in any artifact). Are they implemented by the way? Hope not, so need to fix just this note...

src/config/artifact/validate.rs
```diff
@@ -70,13 +74,9 @@ fn validate_envelope<P: Payload>(
...
     }
-    if env.kind != P::KIND {
-        bail!(Error::new(format!(
-            "Expected a {:?} artifact, got {:?}",
-            P::KIND,
-            env.kind
-        )));
-    }
+    // Name is the whole of dispatch: the caller states which artifact it came
+    // for, and `P` is the schema it decided that name implies. Nothing else in
```
wsnark: This is misleading; caller should never imply anything about payload schema based on the name. A workload should use its name for logging mostly + some defense-in-depth like the name check below.

src/config/rete/model/user.rs
leviathan941: https://github.com/rete-labs/florete/issues/130

src/config/artifact/validate.rs
```diff
@@ -51,15 +51,18 @@ impl ValidatePayload for VertexMgmtPayload {
...
     expected_name: &str,
 ) -> Result<(), Report<Error>> {
     // Fail-closed major.minor gate (see `version`): accept older-or-equal minors
-    // within the known major, reject a different major or a newer minor.
-    version::ensure_supported(&env.schema_version)?;
+    // within the known major, reject a different major or a newer minor. The
+    // envelope and the payload family version independently, so both are gated
+    // — here rather than in each `validate_payload`, so no payload can skip it.
+    version::ensure_supported(&version::ENVELOPE, &env.schema_version)?;
+    version::ensure_supported(&P::FAMILY, env.payload.schema_version())?;
```
wsnark: This is really strange and feels bad that validate_envelope validates payload too. There is `flor agent` use case when it validates the envelope, never parsing the payload - it should be able to validate any payload, even the ones it doesn't know. Also there is `coordinator` which should be able to dispatch artifacts to nodes - without even being able to parse the envelope fully (the frozen core concept). 

The payload itself should be validated by workloads only (as they're the only ones who understand the payloads).

src/config/artifact/validate.rs
wsnark, line 183: Validate should validate `node` name too probably; though need to think about how to do this properly:
1. flor agent knows the node name (its own name) and must validate it
2. workloads normally don't care about node name and cannot validate it
3. coordinator (& `retectl publish` maybe) would like to validate "frozen core" - probably different method (it needs consistency, no version shenanigans, no duplicates on a collection of artifacts - something like this).

So currently we have "vertex-level validate", lacking "agent-level validate". Maybe we can refactor within `flor agent` impl tickets, what do you think? Though splitting validate in payload-agnostic part and payload-aware one makes sense to me as early as possible...

src/config/rete/validate.rs
wsnark: Looks like I confused "8080" and "1080" port numbers, in fact there is no 3 non-unique addrs but only 2, as expected. So eventually it would be better to use visually more distinct port numbers... But here let's keep it as is

src/config/artifact/validate.rs
wsnark, line 190: This lacks usage of new rete.json's trust domain; can be fixed within #43 though, not right here

src/vertex.rs
leviathan941: Agreed. Reworded: the docstring now says the flat `<name>.json` mgmt set is the layout `flor agent` lays down for the workloads it supervises, and that how the set reached the node is none of the vertex's business. The `retectl compile` / `NodeVertexArtifact::path` link is gone.

src/config/artifact/model/vertex.rs
leviathan941: They were in the model, though not in the runtime — C0 warns and ignores a FlorIO channel. Rather than fix just the note, I removed the fields to match the design (`agent.mdx`: "artifacts carry no filesystem references"; C1 mgmt-plane: "no top-level `florio` block, no socket paths"):

- `Adapter::Florio { name, socket: PathBuf }` → `Adapter::Florio { name, vertex }` — the serving link-vertex by name, as in the C1 example `{ "name": "via-link", "type": "florio", "vertex": "public" }`.
- `IoChannel::Florio { socket: PathBuf }` → `IoChannel::Florio {}`, i.e. `{ "kind": "florio" }`.

`PathBuf` no longer appears anywhere under `src/config/artifact/`, so "no paths" is now enforced by the types rather than by prose, and `rejects_the_dropped_path_fields` covers a stray florio `socket` on both the adapter and the io channel.

One doc nit while I was there: ADR-0010 under "`io` direction is intrinsic to the channel kind" still says the kind is "reinforced by whether the entry carries `listen`, `upstream`, or `socket`" — that last one no longer exists. I left the ADR alone; worth an amendment if you agree.

src/config/artifact/validate.rs
leviathan941: Right — reworded. The comment now says `name` is addressing: it states which artifact this is, and checking it is defence-in-depth against a consumer being handed one it did not ask for. It claims nothing about the payload's schema, no consumer may infer one from it, and a workload otherwise uses its name for logging.

src/config/artifact/validate.rs
leviathan941: Agreed — split, and the payload-agnostic half is now the one with no `Payload` bound:

```rust
impl<P> Envelope<P> {
    pub fn validate_claims(
        &self,
        expected_plane: PlaneTag,
        expected_name: &str,
        expected_node: Option<&str>,
    ) -> Result<(), Report<Error>>;
}

impl<P: ValidatePayload> Envelope<P> {
    // claims(P::PLANE, name, None) + the payload family's version gate + payload rules
    pub fn validate(&self, expected_name: &str) -> Result<(), Report<Error>>;
}
```

Because `validate_claims` is unbounded in `P`, `flor agent` can run it over an `Envelope<serde_json::Value>` — envelope parsed, payload opaque — which is exactly the case a relay needs too. The `P::FAMILY` version gate moved out of the claims half deliberately: it is a payload claim, and a consumer that cannot name the family must not be asked to gate it. A test asserts claims still validate over a payload that is not a vertex payload at all, and another that an unsupported payload minor passes the claims check while `validate` rejects it.

src/config/artifact/validate.rs
leviathan941: Agreed on cutting the seam early, so it is in this PR. `node` is now the third argument of `validate_claims`, as an `Option`:

- `None` for a workload — it is handed its config and has no independent notion of the node it runs on (your point 2).
- `Some(node)` for a caller that does know. Today that is the compiler's self-check, which knows both claims and so cannot emit an artifact stamped for the wrong node; `flor agent` passes its own name when it lands (your point 1).

What is still missing is agent-level validation of an artifact *set* and the coordinator's frozen-core consistency check (your point 3) — those need the set as the unit, not a single envelope, so I left them to #42 and noted the gap in the module docs.

leviathan941: Agreed — leaving this as the internal-consistency check it is, with the docstring already saying the trust domain is not verified as *ours* here. Sourcing it from `rete.json` is #43.

src/config/artifact/validate.rs
```diff
@@ -35,11 +51,65 @@ pub trait ValidatePayload: Payload {
...
 
+impl<P> Envelope<P> {
+    /// Validate the envelope's own claims, learning nothing about the payload:
+    /// the envelope contract's version, the plane this artifact was expected on,
+    /// the name it was fetched under, and — for a caller that knows it — the
+    /// node it was projected for.
+    ///
+    /// Deliberately unbounded in `P`, so a consumer that cannot parse the
+    /// payload (the agent handing an artifact down to a workload, a relay moving
+    /// one along) can still check everything the frozen core claims.
```
wsnark: Relay needs different validation function - it shouldn't check schema_version, because it must relay incompatible versions by design; let's create a ticket for it, but don't pretend we already have it covered.

src/config/compile.rs
```diff
@@ -88,9 +88,14 @@ pub fn compile(
...
         let artifact = vertex::project(&plan, node, node_plan, opts)?;
+        // The compiler is the one consumer that knows both claims, so it checks
+        // the `node` too: an artifact stamped for the wrong node must never
+        // leave here. (A workload cannot make this check — it is handed its
+        // config and has no independent notion of its node.)
         artifact
             .envelope
-            .validate(&artifact.vertex_name)
+            .validate_claims(PlaneTag::Mgmt, &artifact.vertex_name, Some(&artifact.node))
+            .and_then(|()| artifact.envelope.validate(&artifact.vertex_name))
```
wsnark: This isn't very efficient - `validate_claims` is called twice; probably this shows that validation API isn't well-designed... It has a lot of different use cases:
- validate at retectl compile 
- validate at retectl publish
- validate at flor agent
- validate at flor workload (vertex)
- validate at flor coordinator as consumer (its status is unclear - sometimes it is treated as "third-party", not part of flor; sometimes not...e.g. from validation PoV now it is external, while it will be built from the same codebase now)
- validate a published set at flor coordinator as relay ("frozen core" only)

Let's create a ticket for it. Or maybe we should just improve this validation API gradually, when implementing these parts?

wsnark (approved):

src/config/compile.rs
leviathan941: Yes, I think we must implement things when we really need them and don't try to foresee them.

src/config/artifact/validate.rs
leviathan941: This method is for `Envelope`. Each envelope has `schema_version` field, it's declared in the struct.

src/config/compile.rs
leviathan941: Changed API. Now we have separate struct `Expect`, where a client set what it knows. It's extensible for new clients.

src/config/artifact/validate.rs
wsnark: Yep, but "frozen core"/"routing core" is different -- https://florete.tech/docs/implementation/adr/0012-version-artifact-contract/#the-routing-core-the-defined-frozen-relay-contract  - and the difference is in "schema_version" mainly - relay shouldn't check it, because it must serve artifacts of versions it doesn't understand (that's why this "routing core" was invented, by the way). So pleas fix this wording - it is OK that this func is for Envelope only - and then it shouldn't mention "relay" and "frozen core" - the Envelope itself is fully checked at producer and consumer sides, but never at the relay.

src/config/artifact/validate.rs
```diff
@@ -31,11 +52,93 @@ pub trait ValidatePayload: Payload {
...
+    /// and whatever the caller [expects](Expect) of it.
+    ///
+    /// Deliberately unbounded in `P`, so a consumer that cannot parse the
+    /// payload can still check everything the frozen core claims. `plane` is
+    /// an argument for the same reason: a `P` that states nothing cannot supply
+    /// it, and it must never be silently skipped. A caller that owns the payload
+    /// type uses [`validate`](Envelope::validate), which takes it from `P`.
+    pub fn validate_claims(
+        &self,
+        expected_plane: PlaneTag,
```
wsnark: why expected_plane isn't part of Expect then? looks strange. If it is mandatory - OK, same like `name` in Expect?

src/config/artifact/validate.rs
```diff
@@ -31,11 +52,93 @@ pub trait ValidatePayload: Payload {
...
+        expect: Expect<'_>,
+    ) -> Result<(), Report<Error>> {
+        // Fail-closed major.minor gate (see `version`): accept older-or-equal
+        // minors within the known major, reject a different major or a newer
+        // minor. This is the envelope's ladder only; the payload family rides
+        // its own, gated by whoever parses the payload.
+        version::ensure_supported(&version::ENVELOPE, &self.schema_version)?;
```
wsnark: version check should be optional in Expect, I suppose - the frozen relay contract isn't enforcing it... Though it should be opt-out, not opt-in probably

src/config/artifact/validate.rs
wsnark: Well, if we improve Expect to allow opting-out of a version check, we can cover "frozen core" case too; docstring wording still needs update to reflect this

wsnark: I think I was partially wrong with previous comments. The situation is really a bit more complicated, let's look deeper.
I have specified the high-level requirements in #66. I think consequences for Expect API are following:
- let all fields be optional
- let schema_version check be optional as well

Probably we shouldn't add more checks now - only when implementing corresponding consumers of the API, right?

src/config/artifact/validate.rs
leviathan941: Added, now it's a part of Expect.

src/config/artifact/validate.rs
leviathan941: Yes, you're right, now it's opt-out.

src/config/artifact/validate.rs
leviathan941: I think I get we you are talking about. Look at the latest version of Expect.

wsnark:
New `Expect` looks good! Let's merge it and evolve the API as consumers are implemented.

PR #70, opened 2026-07-31
feat: migrate our logger to tracing (#67)

### LINKED ISSUE #67: Adopt `tracing` as the logging facade, with the first instrumented codepath
Design: [C0 Observability § Facade](https://florete.tech/docs/implementation/c0-tended-tunnels/observability#facade-tracing-adopted-in-c0-and-used-sparingly) (rete-labs/florete#122).

Move flor's logging from `log` + `env_logger` onto `tracing`, and instrument one codepath to exercise it. The payoff is span context: events emitted inside an operation become attributable to it, including events from code that knows nothing about spans — which is what makes concurrent codepaths readable.

## What

- Replace `logging::logger` with a `tracing` subscriber, preserving the current line format, the `Config` surface, and `RUST_LOG` behaviour. Existing `log::` call sites and third-party crates must keep working without being migrated.
- Instrument the SOCKS5 inbound as the first codepath, so its concurrent connections are distinguishable in the log.
- Fix the pre-existing bug where output is colorized when stderr is not a terminal.

Span granularity and level are constrained by the design section above.

One known trap, since it costs time to find: records arriving through the `log` bridge carry fixed metadata — the component becomes literally `log`, with the real one in a `log.target` field. `tracing_log::NormalizeEvent` recovers it; without that, every bridged line prints the wrong component.

## Acceptance

- Line format unchanged for events outside a span; span context rendered before the message.
- `RUST_LOG` works as before, both global and per-component.
- No ANSI escapes in output when stderr is not a terminal.
- Every log line from a proxied connection is attributable to that connection, at any filter level that shows the line at all.
- Existing logger unit tests survive the change.
- Full suite green; clippy and fmt clean.

## Follow-ups

- #68 — instrument the remaining codepaths.
- #69 — flood-safe line Layer (depends on #68).

wsnark:
This is initial commit with single instrumentation site (socks5 inbound) that demonstrates new tracing capability.

Closes #67

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ version = "0.1.0"
 async-trait = "0.1"
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
 clap = { version = "4", features = ["derive"] }
-env_logger = "0.11"
 error-stack = "0.7"
 fast-socks5 = "1.0"
 fundle = "0.3"
@@ -25,12 +24,16 @@ thiserror = "2.0"
 tokio = { version = "1", features = ["full"] }
 
 # Using "ring" backend across all crypto libs, as quinn uses it by default
+anstyle = "1.0"
+pem = "3"
 quinn = { version = "0.11" }
-rcgen = { version = "0.14", default-features = false, features = ["ring", "pem", "x509-parser"] }
+rcgen = { version = "0.14", default-features = false, features = ["pem", "ring", "x509-parser"] }
 rustls = { version = "0.23", default-features = false, features = ["ring", "std"] }
 spiffe = { version = "0.15", default-features = false, features = ["x509"] }
-pem = "3"
 time = { version = "0.3", default-features = false, features = ["std"] }
+tracing = "0.1"
+tracing-log = "0.2"
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 x509-parser = { version = "0.18", features = ["verify"] }
 
 [dev-dependencies]
--- a/src/logging/logger.rs
+++ b/src/logging/logger.rs
@@ -1,14 +1,38 @@
 // Copyright (C) 2026 ReteLabs LLC.
 // Licensed under Apache-2.0 or MIT at your option.
 
-use std::{borrow::Cow, io::Write};
+//! The process-wide logger: a `tracing` subscriber rendering one line per event.
+//!
+//! Events reach it from both facades — `tracing::` macros and, through
+//! `LogTracer`, the `log::` macros used by flor and its dependencies — so a
+//! single filter and format govern everything. Enclosing spans are rendered
+//! before the message, which is how an event emitted deep inside a connection
+//! handler carries that connection's identity without being passed it.
+//!
+//! Call [`init`] or [`init_with_config`] once per process, before anything logs.
+//! `RUST_LOG` overrides the [`Config`] filters and accepts the usual
+//! `target=level` syntax.
 
+use std::borrow::Cow;
+use std::fmt;
+use std::io::IsTerminal;
+
+use anstyle::{AnsiColor, Effects, Style};
 use chrono::Local;
-use env_logger::fmt::style::{self, Style};
 use error_stack::{Report, ResultExt};
-use log::LevelFilter;
+use tracing::{Event, Subscriber};
+use tracing_log::NormalizeEvent;
+use tracing_subscriber::fmt::format::Writer;
+use tracing_subscriber::fmt::{FmtContext, FormatEvent, FormatFields, FormattedFields};
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::registry::LookupSpan;
+use tracing_subscriber::util::SubscriberInitExt;
 
 use crate::logging::Error;
+
+/// Severity threshold for the logger, re-exported so call sites depend on this
+/// module's vocabulary rather than on whichever facade backs it.
+pub use tracing::level_filters::LevelFilter;
 
 const SHORTENED_TARGET_MAX_LEN: usize = 20;
 
@@ -81,7 +105,7 @@ impl Config {
 impl Default for Config {
     fn default() -> Self {
         Self {
-            global_log_filter: LevelFilter::Info,
+            global_log_filter: LevelFilter::INFO,
             modules_log_filter: vec![],
             include_time: true,
             include_date: false,
@@ -99,11 +123,10 @@ impl Default for Config {
 ///
 /// ### Examples
 /// ```
-/// use log::LevelFilter;
-/// use flor::logging::logger;
+/// use flor::logging::logger::{self, LevelFilter};
 ///
-/// logger::init(LevelFilter::Info).unwrap();
-/// log::info!("Logger is initialized!");
+/// logger::init(LevelFilter::INFO).unwrap();
+/// tracing::info!("Logger is initialized!");
 /// ```
 pub fn init(log_level: LevelFilter) -> Result<(), Report<Error>> {
     init_with_config(&Config::new().global_log_filter(log_level))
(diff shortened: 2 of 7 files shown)


wsnark:
What changed in logger output:
1. Instrumented span adds context to each log line in curly brackets (see example below)
2. error_stack's color mode is changed:
   - debug build + terminal output: Color mode (+ tracing_subscriber allowing it, i.e. not escaping)
   - everything else: No color mode (+ tracing_subscriber escaping any ansi-styling in the messages - a security feature we keep for release builds)

What changed in API:
`log::LevelFilter` is replaced with our own `logger::LevelFilter`, which is based on tracing-provided one instead of log-provided. Mostly a cometic change. Though call sites need an update; also level notation changed from camel-case to upper-case: `LevelFilter::Info` to `LevelFilter::INFO` (that's due to tracing's LevelFilter).

Everything else kept the same.

How to test:
```bash
cd <flor-repo>
# prepare dev env (in case it is missing or from different branch)
rm ./.flor-dev -rf
./scripts/dev-bootstrap.sh
# run a vertex 
RUST_LOG=trace FLOR_HOME=./.flor-dev/ cargo run --bin flor -- vertex run --rete alpha --name flor
# trigger new tracing site in another terminal:
curl https://ttt.com --socks5-hostname 127.0.0.1:1080
# trigger error stack
RUST_LOG=trace FLOR_HOME=./.flor-dev/ cargo run --bin flor -- vertex run --rete alpha --name flor2
```

Screenshot of instrumented span:
Screenshot From 2026-07-31 20-31-22

Screenshot of colorized error-stack:
Screenshot From 2026-07-31 21-38-24

src/logging/logger.rs
```diff
@@ -122,129 +145,222 @@ pub fn init(log_level: LevelFilter) -> Result<(), Report<Error>> {
...
+    // (hostnames, SPIFFE IDs, remote error strings), so a release build always
+    // keeps it. The cost is that a styled `Report` arrives mangled rather than
+    // colored, since its escapes travel inside the message.
+    //
+    // Debug builds trade that away: a developer reading an error stack in a
+    // terminal benefits from the color, and the peers involved are their own. The
+    // relaxation is scoped as narrowly as it can be — debug build *and* a terminal
+    // sink — so a redirected dev log is still sanitized.
+    // Release's log is always sanitized.
+    let styled_reports = ansi && cfg!(debug_assertions);
```
leviathan941: I think this should be configurable by `Config` and logger must be as stupid as possible.
`is_terminal && debug` looks like a default value, but this should not be the area of responsibility of the logger itself.

Wdyt?

src/logging/logger.rs
```diff
@@ -122,129 +145,222 @@ pub fn init(log_level: LevelFilter) -> Result<(), Report<Error>> {
...
+        .change_context(Error("Failed to set logger".into()))
+}
+
+/// Maps the `Config` filters onto an `EnvFilter`, with `RUST_LOG` taking precedence.
+fn build_filter(config: &Config) -> tracing_subscriber::EnvFilter {
+    if let Ok(env) = std::env::var("RUST_LOG")
+        && !env.is_empty()
+    {
+        return tracing_subscriber::EnvFilter::new(env);
+    }
```
leviathan941: Have you tried `EnvFilter::try_from_default_env()`: [try_from_default_env](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#method.try_from_default_env)?

src/logging/logger.rs
```diff
@@ -122,129 +145,222 @@ pub fn init(log_level: LevelFilter) -> Result<(), Report<Error>> {
...
+
+    let mut directives = vec![config.global_log_filter.to_string()];
+    for (module, level) in &config.modules_log_filter {
+        directives.push(format!("{module}={level}"));
+    }
+    tracing_subscriber::EnvFilter::new(directives.join(","))
```
leviathan941: It looks like `filter::Builder` can be used here.
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.Builder.html#method.parse.

src/northbound/inbound/socks5.rs
```diff
@@ -159,28 +160,37 @@ async fn bind_listeners(
...
     let mut tasks = JoinSet::new();
     loop {
         tokio::select! {
             result = listener.accept() => match result {
                 Ok((stream, peer_addr)) => {
-                    log::debug!(target: LOG_TARGET,
-                        "Accepted connection from {peer_addr} for principal '{principal}'");
+                    // `error_span!` so the context survives any filter level: an
+                    // `info_span!` vanishes under `RUST_LOG=warn`, taking the
+                    // attribution of the WARN lines inside it with it.
```
leviathan941: Is this comment really necessary here? It looks like a doc for the. logger, not for the socks5.

src/northbound/inbound/socks5.rs
```diff
@@ -231,12 +241,16 @@ async fn handle_socks5(
...
+    tracing::Span::current().record("target", tracing::field::display(&host));
+
     // Step 4: Resolve the hostname to a dial target against the caller's trust
     // domain (inbound builds services only; see ADR-0007).
     let target = match Dialable::resolve(&host, caller.spiffe_id().trust_domain()) {
         Ok(target) => target,
         Err(e) => {
-            log::debug!(target: LOG_TARGET, "Cannot resolve target '{host}': {e:?}");
+            // Rejected user input, not a fault: the span already carries the
+            // target, and `{e}` reports the cause without an error-stack trace.
```
leviathan941: I don't think this comment is useful. Host is added to the span by 4 lines of code above.

Cargo.toml
```diff
@@ -32,6 +31,10 @@ spiffe = { version = "0.15", default-features = false, features = ["x509"] }
 pem = "3"
 time = { version = "0.3", default-features = false, features = ["std"] }
 x509-parser = { version = "0.18", features = ["verify"] }
+tracing = "0.1"
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+tracing-log = "0.2"
+anstyle = "1.0"
```
leviathan941: alphabetical order is violated here

leviathan941 (changes requested):

src/logging/logger.rs
wsnark: That is more or less equivalent - the` try_from_default_env` just errors on invalid directives too, while `new` uses `parse_lossy`, ignoring invalid ones. So probably we should decide whether we reject whole RUST_LOG if invalid directive found, or ignore the invalid one... Well, silently ignoring may be bad UX - user will think that he enabled debug output for some component, but in fact a typo prevented it, and silently ignored. So let's use strict parsing of RUST_LOG

wsnark: On the other hand, a typo in a component name cannot be detected here. So RUST_LOG is by design not very safe...

src/logging/logger.rs
wsnark: `new` uses similar `parse_lossy` under the hood, which is equivalent because our own generated directives are always correct

src/northbound/inbound/socks5.rs
wsnark: Agreed. This rule is documented at https://florete.tech/docs/contributing/development/conventions/#logging-an-error-stack-for-faults-one-line-for-expected-outcomes (Span level gates context)

src/northbound/inbound/socks5.rs
wsnark: Agreed. The rule is documented at https://florete.tech/docs/contributing/development/conventions/#logging-an-error-stack-for-faults-one-line-for-expected-outcomes (+ Don't interpolate what the span already carries)

Cargo.toml
wsnark: Good catch. We don't have auto-fmt for Cargo.toml, unfortunately...

src/logging/logger.rs
wsnark: Well, our logger is fixed to stderr currently, and I think that's fine - until we add support for structured logging (like directly into journald). Current design for flor agent is to capture stderr of all supervised workloads, marker it with workload name and forward to its own stderr, which will end up in journald or log file (on macOS) - ensured by service wrapper.

`is_terminal` shouldn't be a config - it is a property of log sink (stderr that isn't configurable now).

Second, opt-out of terminal escaping shouldn't be an option of the logger I think - it is a security feature that must be enabled. We relax it for debug builds only.. So I think it is proper place to check.

src/logging/logger.rs
wsnark: Claude found a good improvement: we should parse this strictly (via `try_new`) so invalid programmer-supplied directives are an error

src/logging/logger.rs
wsnark: Kept current code because it is good: `new` uses `parse_lossy` which prints a warning to stderr if it finds something bad: 
image

But parsing cannot ensure valid RUST_LOG - as it is string-based (e.g. a typo in a component name). So we shouldn't error-out on a bad directive in RUST_LOG.

`try_from_default_env` doesn't distinguish between RUST_LOG not set or containing bad directive, so it isn't useful for us.

Cargo.toml
leviathan941: Let's create a ticket for that.

src/logging/logger.rs
leviathan941: Ok, let it be.

src/logging/logger.rs
```diff
@@ -122,129 +145,233 @@ pub fn init(log_level: LevelFilter) -> Result<(), Report<Error>> {
...
+    // (hostnames, SPIFFE IDs, remote error strings), so a release build always
+    // keeps it. The cost is that a styled `Report` arrives mangled rather than
+    // colored, since its escapes travel inside the message.
+    //
+    // Debug builds trade that away: a developer reading an error stack in a
+    // terminal benefits from the color, and the peers involved are their own. The
+    // relaxation is scoped as narrowly as it can be — debug build *and* a terminal
+    // sink — so a redirected dev log is still sanitized.
+    // Release's log is always sanitized.
+    let styled_reports = ansi && cfg!(debug_assertions);
```
leviathan941: One more thing.
We already have is_debug definition in retectl: https://github.com/rete-labs/flor/blob/main/src/bin/retectl.rs#L133.
I think we should use only one source of truth for that all over the project. So, it looks like it's better to move it into util and use here and there.

But, I don't insist to do this in this PR.

leviathan941 (approved):

src/logging/logger.rs
```diff
@@ -122,129 +147,247 @@ pub fn init(log_level: LevelFilter) -> Result<(), Report<Error>> {
...
+/// This behaviour is justified because we cannot ensure that `RUST_LOG` is fully
+/// correct: a typo in a component name cannot be detected.
+///
+/// On the other hand, `Config` directives are programmer input, so a malformed one
+/// is a bug and fails initialization instead of being skipped.
+fn build_filter(config: &Config) -> Result<tracing_subscriber::EnvFilter, Report<Error>> {
+    if let Ok(env) = std::env::var("RUST_LOG")
+        && !env.is_empty()
+    {
+        return Ok(tracing_subscriber::EnvFilter::new(env));
```
leviathan941: This changes the established RUST_LOG composition. Previously the builder first installed the Config global filter and then added environment directives. With a component-only value such as RUST_LOG=bridge_mod=trace, unrelated INFO events used to remain visible; EnvFilter::new instead supplies an ERROR default for every unmatched component. It also drops the old env_logger regex form. Please preserve the old global and per-component semantics, or explicitly narrow the compatibility requirement.

src/northbound/inbound/socks5.rs
```diff
@@ -159,28 +160,34 @@ async fn bind_listeners(
...
-    let principal = caller.spiffe_id().clone();
+    let initiator = caller.spiffe_id().clone();
     let mut tasks = JoinSet::new();
     loop {
         tokio::select! {
             result = listener.accept() => match result {
                 Ok((stream, peer_addr)) => {
-                    log::debug!(target: LOG_TARGET,
-                        "Accepted connection from {peer_addr} for principal '{principal}'");
+                    let span = tracing::error_span!(
```
leviathan941, line 169: An error_span only contributes context while its own target is enabled. For example, RUST_LOG=off,quinn=trace can show a third-party event inside this connection while the socks5_inbound span is disabled, leaving the line un-attributed. This is precisely the case tracing is meant to improve. Please keep operation spans enabled independently of event filtering, or add an equivalent base directive.

src/main.rs
```diff
@@ -105,7 +105,8 @@ fn main() {
         Cmd::Vertex {
             action: VertexAction::Run(vertex_args),
         } => {
-            logging::logger::init(log::LevelFilter::Info).expect("Failed to initialize logger");
+            logging::logger::init(logging::logger::LevelFilter::INFO)
```
leviathan941, line 108: Report color mode is configured only through logger::init, but this runs only for Vertex::Run. The flor --verbose id keygen error path does not initialize the logger and still emits ANSI escapes when stderr is redirected. Please configure the error-stack color policy before command dispatch, independently of subscriber initialization.

src/logging/logger.rs
```diff
@@ -122,129 +147,247 @@ pub fn init(log_level: LevelFilter) -> Result<(), Report<Error>> {
...
+        .event_format(format);
+
+    tracing_subscriber::registry()
+        .with(build_filter(config)?)
+        .with(fmt_layer)
+        .try_init()
+        .change_context(Error("Failed to set logger".into()))?;
+
+    if unsanitized {
+        tracing::warn!(
```
leviathan941, line 218: The comment says the unsanitized mode is announced, but this warning is subject to the normal subscriber filter. For example, FLOR_LOG_UNSANITIZED=1 with RUST_LOG=probe=warn enables terminal escape pass-through while filtering out this flor-target warning. Please emit this static security notice directly to stderr or otherwise bypass normal event filtering.

src/northbound/inbound/socks5.rs
```diff
@@ -159,28 +160,34 @@ async fn bind_listeners(
...
-                    tasks.spawn(async move {
-                        if let Err(e) = handle_socks5(stream, backend, caller).await {
-                            log::warn!(target: LOG_TARGET,
-                                "Connection from {peer_addr} via SOCKS5 for principal \
-                                 '{principal}' error: {e:?}");
+                    tasks.spawn(
+                        async move {
+                            log::debug!(target: LOG_TARGET, "Accepted connection");
+                            if let Err(e) = handle_socks5(stream, backend, caller).await {
+                                log::warn!(target: LOG_TARGET, "SOCKS5 connection error: {e:?}");
```
leviathan941, line 182: The connection-error path formats a Report with Debug, which embeds newlines. FlorFormat writes the conn span prefix only once, so continuation lines such as the error-stack tree are bare stderr lines once the agent captures and re-emits them. That breaks the requirement that every line from a proxied connection remains attributable. Please use a one-line event for expected peer failures, or prefix every physical line when a full stack is required.

src/logging/logger.rs
wsnark: Agreed it is worth unifying, but the root is deeper than one predicate: \`retectl\` prints the compact chain *and*, in debug builds, the full stack — the same error twice — while \`flor id\` selects between them with \`--verbose\`. Those are alternatives, not layers, and neither path initializes a logger at all, so the error-stack color policy is unset there too.

Also note the second use site is gone: now the logger keys off \`FLOR_LOG_UNSANITIZED\` rather than the build profile, so there is nothing left in it to share.

Filed as rete-labs/florete#135 (design) — including the question of whether console utilities should write to the system log at all. Out of scope here.

src/northbound/inbound/socks5.rs
wsnark: ERROR is already the ceiling: a span is gated by its own metadata, and no level is more permissive, so this span survives every filter except one that turns the flor target off outright. `RUST_LOG=off,quinn=trace` is that case — it says "silence flor" and gets silence, including flor's spans.

Keeping spans enabled independently of event filtering is not expressible with a single global `EnvFilter`; it needs per-layer filters, which is a disproportionate cost for a directive that asks for this outcome. Not changing.

src/logging/logger.rs
wsnark: The warning is a log line and should obey the log filter: an operator who sets `RUST_LOG=error` has asked not to see warnings, and a notice that bypasses filtering to reach stderr would also have to reproduce the timestamp, level and target formatting by hand, diverging from every other line.

The case in the example is a real bug, but a different one — a component-only `RUST_LOG` was silencing every other target instead of leaving it at the `Config` level. Fixed separately in this PR; with that, `FLOR_LOG_UNSANITIZED=1 RUST_LOG=probe=warn` shows the notice.

Cargo.toml
wsnark: Filed as #71 — manifest formatting enforced in CI, not only ordering.

src/main.rs
wsnark: Correct as a finding, but out of scope here: #67 migrates the logger from `env_logger` to `tracing`, and this path never initializes a logger at all — the escapes come from `error_stack`'s default `ColorMode::Emphasis`, which predates this PR.

Fixing it now would also presume an answer to a question that is still open: whether console utilities should print error stacks at all. If they do not — leaving stacks to the logger and printing only the compact chain — then no console path renders a `Report` and there is no color mode to set. Captured on rete-labs/florete#135, which owns that decision.

src/northbound/inbound/socks5.rs
wsnark: Keeping `{e:?}` here. This is not the "user mistyped a hostname" case the one-line rule covers: reaching it means the SOCKS5 library or our own backend failed, i.e. a fault that should be very rare — exactly where the causal chain and source locations are the point.

The premise about newlines does not hold either, and it is worth writing down why. The subscriber formats an entire event into one buffer and emits it with a single `write` on an unbuffered stderr, so a multi-line message is one write syscall, serialized against the other threads in the process — the writer never splits at `\n`.

What is real is the atomic-write floor on the `shared` sink class, where siblings write to one tty or pipe: 4096 bytes on Linux but 512 on macOS, and a rendered stack can exceed that. Captured in [C0 Observability § The log stream](https://florete.tech/docs/implementation/c0-tended-tunnels/observability#the-log-stream), with #72 to split an oversized report into several complete messages.

src/logging/logger.rs
wsnark: Good catch! Fixed now

leviathan941 (approved):

